PR #11, opened 2022-08-10
update-mui-design-system

ficast:
image

### DIFF
--- a/packages/design-system-mui/components/ButtonWithLoading/styled.tsx
+++ b/packages/design-system-mui/components/ButtonWithLoading/styled.tsx
@@ -1,13 +1,6 @@
 import { styled } from '@mui/material/styles'
-import {
-  Button as MUIButton,
-  ButtonProps,
-  CircularProgress as MuiCircularProgress,
-  CircularProgressProps,
-} from '@mui/material'
+import { CircularProgress as MuiCircularProgress, CircularProgressProps } from '@mui/material'
 import { FC } from 'react'
-
-export const Button = styled(MUIButton)(({ theme }) => ({})) as unknown as FC<ButtonProps>
 
 export const CircularProgress = styled(MuiCircularProgress)(({ theme }) => ({
   position: 'absolute',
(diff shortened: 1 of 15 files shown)


packages/design-system-mui/components/ButtonWithLoading/styled.tsx
```diff
@@ -1,6 +1,18 @@
...
+  ButtonProps,
+  CircularProgress as MuiCircularProgress,
+  CircularProgressProps,
+} from '@mui/material'
+import { FC } from 'react'
 
-export const Button = styled(MUIButton)(({ theme }) => ({
-  background: theme.palette.primary.main,
-}))
+export const Button = styled(MUIButton)(({ theme }) => ({})) as unknown as FC<ButtonProps>
```
anicioalexandre: no really need to define it here if we are not customizing it

packages/design-system-mui/components/ButtonWithLoading/styled.tsx
```diff
@@ -1,6 +1,18 @@
...
+
+export const CircularProgress = styled(MuiCircularProgress)(({ theme }) => ({
+  position: 'absolute',
```
anicioalexandre, line 6: wouldn't be better if we show the loading state besides the children text? Not sure how the design looks like but seems better than overlapping the content

packages/design-system-mui/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@emotion/react": "^11.10.0",
     "@emotion/styled": "^11.10.0",
+    "@mui/icons-material": "^5.8.4",
```
anicioalexandre, line 24: let's not add this until we need

ficast: I am using that

packages/design-system-mui/components/ButtonWithLoading/styled.tsx
ficast: you can always have your own loadingChildren to overwrite

anicioalexandre (approved):

PR #18, opened 2022-08-24
BA-899: Replace formik with react hook form

ghost:
on @baseapp-frontend/core:

- replace formik with react-hook-form
- make sure the hooks still working as expected
    - it's expected that, with the migration, projects that uses the core package hooks stop working, so let's do the best implementation we can not worrying  much about that because the projecst that uses the @baseapp-frontend/core will have to adapt no matter what

- add examples of the core implementations inside apps/docs if possible
    - ex: on pages/login : example of useLogin() usage 

I also replaced formik in the MUI components because those were going to break otherwise.


ficast (approved):

dpTsl (approved):

packages/core/src/auth/resetPassword.tsx
```diff
@@ -1,7 +1,9 @@
 import { IResetPassword } from './types'
 import * as Yup from 'yup'
+import { yupResolver } from '@hookform/resolvers/yup'
 import { axios, useMutation } from '../api'
-import { useFormik } from 'formik'
+import { useForm } from 'react-hook-form'
```
anicioalexandre, line 4: I know you are going to hate me lol, but could we organize the imports?
```
// lib imports:
import * as Yup from 'yup'
...
< ---- space ---- >
// relative imports:
import { axios, useMutation } from '../api'
...
```

packages/design-system-mui/components/withController/index.tsx
```diff
@@ -0,0 +1,45 @@
+import { Controller } from 'react-hook-form'
+import { IControlledComponentProps } from './types'
+
+function withController(Component: any) {
```
anicioalexandre: do u think we could move that `withController` to another package? Maybe `core` or we could create a `utils`

It being on the `design-system-mui` does not make much sense to me, since we could use that HOC for other purposes that does not include that design-system-mui itself (like, let's say I wanna create my own components using the react-hook-form but I wanna take advantage of the `withController`)

anicioalexandre (changes requested):

packages/core/src/auth/resetPassword.tsx
ghost: LOL, sure. I think we should introduce an autosort to not have to do this anymore. We started using a prettier plugin on FAVIO that works pretty well for that. I'll bring it up in the next frontend meeting.

packages/design-system-mui/components/withController/index.tsx
ghost: You are right, we could use this with other UI frameworks (hopefully). I'll move it!

ghost: Done, but now `design-system-mui` depends on `core`, which is probably ok, but not sure if we really want that.

Janekk (approved):

Janekk:
unrelated to this PR, noticed we repeat this prop type definition with different core hooks:

```export function useSignUp({
  validationSchema = signUpValidationSchema,
  defaultValues = defaultInitialValues,
  onError,
  onSuccess,
}: {
  validationSchema?: any
  defaultValues?: any
  // eslint-disable-next-line @typescript-eslint/ban-types
  onError?: Function
  // eslint-disable-next-line @typescript-eslint/ban-types
  onSuccess?: Function
})
```
we prolly should extract and reuse

packages/core/src/auth/resetPassword.tsx
Janekk: Seems like this prettier plugin would do what we want https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports

anicioalexandre: I lik

> unrelated to this PR, noticed we repeat this prop type definition with different core hooks:
> 
> ```
>   validationSchema = signUpValidationSchema,
>   defaultValues = defaultInitialValues,
>   onError,
>   onSuccess,
> }: {
>   validationSchema?: any
>   defaultValues?: any
>   // eslint-disable-next-line @typescript-eslint/ban-types
>   onError?: Function
>   // eslint-disable-next-line @typescript-eslint/ban-types
>   onSuccess?: Function
> })
> ```
> 
> we prolly should extract and reuse

good catch, that's should be fixed on https://silverlogic.atlassian.net/browse/BA-902 @Janekk

anicioalexandre: that's coool

anicioalexandre:
> unrelated to this PR, noticed we repeat this prop type definition with different core hooks:
> 
> ```
>   validationSchema = signUpValidationSchema,
>   defaultValues = defaultInitialValues,
>   onError,
>   onSuccess,
> }: {
>   validationSchema?: any
>   defaultValues?: any
>   // eslint-disable-next-line @typescript-eslint/ban-types
>   onError?: Function
>   // eslint-disable-next-line @typescript-eslint/ban-types
>   onSuccess?: Function
> })
> ```
> 
> we prolly should extract and reuse

good catch, that's should be fixed on https://silverlogic.atlassian.net/browse/BA-902 @Janekk

packages/design-system-mui/components/withController/index.tsx
anicioalexandre: I think it's fine for now, for the future we're gonna have rethink core any way

anicioalexandre (approved):

PR #21, opened 2022-09-30
BA-926: Multi factor authentication core

Janekk:
* hooks for 2FA log in and accessing MFA endopints for activating and deactivating MFA methods


packages/core/src/utils/functionUtils.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function F(fct: Function | undefined) {
```
ghost: Cool that we abstracted this, but I'm not sure if we really need it. Couldn't we just do `fct?.()` whenever we can have a callable or undefined? Or is it because of the `Function` type?

packages/core/src/mfa/api.ts
```diff
@@ -0,0 +1,60 @@
+import { axios } from '../api'
+import { AxiosResponse } from 'axios'
+import { IPaginatedResponse } from '../types'
+import { ILoginMfaRequest, ILoginResponse } from '../auth/types'
+
+const baseUrl = '/auth'
+
+export interface IMfaActiveMethod {
```
ghost: I think we should move these interfaces to `types.d.ts` so we have every type definition there.

packages/core/src/mfa/types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  smsTwilio = 'sms_twilio',
+}
+
+export declare type MfaMethod = `${MfaMethodEnum}`
+
+export type MfaActivationEmailRequest = {
+  email: string
+}
+
+export type MfaActivationAppRequest = void
```
ghost: I don't see where we are using this type (and we probably don't need it if it's just `void`). Remove?

packages/core/src/utils/functionUtils.ts
jktsl: yea, just a wrapper for the repeating code we had there, i guess we don't need this additional type check if already type already defined with TS.

jktsl:
all the repositories involved:

* https://github.com/silverlogic/django-trench/pull/1 - forked and updated project that will be used as Django app that can be added to BE Baseapp
* https://bitbucket.org/silverlogic/baseapp-django-v2/pull-requests/151/ba-927-mfa-endpoints - BE Baseapp using Dj trench
* https://github.com/silverlogic/baseapp-frontend/pull/21 - MFA core FE
* https://bitbucket.org/silverlogic/baseapp-nextjs-mui-template/pull-requests/16 - MFA MUI template

DEMO: https://www.loom.com/share/e0b80ae5b3dc42a0a8962f04cbf08272

anicioalexandre (approved):

anicioalexandre:
@Janekk a little test failing

anicioalexandre:
we need to refactor this core package to avoid putting everything in there XD
we're planning to do that on the retreat when we'll have time at the Frontend CoP

ghost (approved):

PR #22, opened 2022-12-02
BA-965: Update the API path for second step of 2FA request (code)

Janekk:


pablogadhi (approved):
We need to generate the changeset @Janekk

PR #41, opened 2023-06-26
(GGL-30) customizable hooks

anicioalexandre:
* authentication package update - `v1.0.1`
  - Allows more hook's options to be overridable.

* utils package update - `v1.0.1`
  - Exports `createAxiosInstance` and `getApiErrorMessage` functions.

### DIFF
--- a/packages/authentication/modules/access/useLogout/types.d.ts
+++ b/packages/authentication/modules/access/useLogout/types.d.ts
@@ -1,0 +1,3 @@
+import { ICookieName } from '../../../types/auth'
+
+export interface IUseLogout extends ICookieName {}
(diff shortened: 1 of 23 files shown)


packages/authentication/modules/access/useLogout/types.d.ts
```diff
@@ -0,0 +1,3 @@
+import { ICookieName } from '../../../types/auth'
+
+export interface IUseLogout extends ICookieName {}
```
deboracosilveira, line 3: nothing else to add here? is it just for using another name?

deboracosilveira (approved):

packages/authentication/modules/access/useLogout/types.d.ts
anicioalexandre: yee, that one only receives `cookieName` as prop right now, I kept the interface name just in case we wanna add more into it in the future

julianmn1 (approved):

PR #51, opened 2023-09-06
BA-1136: JWT Flow

MattGalvis:
Acceptance Criteria 

implement JWT login on the baseapp-frontend-template
remove default login flow that we have no from MUI BA-FE-Template (not modules)
make backend-template default authentication method to be JWT
write react hook that gets user info from JWT
update FE templates to use the new hook by default
implement token refresh logic 

Backend template PR: https://bitbucket.org/silverlogic/baseapp-django-v3/pull-requests/3
Backend fixes PR: https://github.com/silverlogic/baseapp-backend/pull/31
DEMO: https://www.loom.com/share/2f4e20eadcd6419e942c429d26a5b299?sid=ba862366-c2c0-468a-8f42-bc0d8c7e5c98


yarn.lock
pablogadhi, line 1: I need to clean the yarn.lock up

packages/authentication/services/auth.ts
```diff
@@ -10,6 +10,10 @@ import {
 
 export default class AuthApi {
   static login({ email, password }: ILoginRequest): Promise<LoginResponse> {
```
Janekk, line 12: I wonder if we should assume a common api path prefix for auth related things on the FE, like we to on auth test project: https://github.com/silverlogic/baseapp-backend/blob/b5bbe8b6baeb4534c1168428c4848d680de52558/baseapp-auth/testproject/urls.py#L13

so that we could have AuthApi:

export default class AuthApi {
  static jwtLogin({ email, password }: ILoginRequest): Promise<LoginResponse> {
    return axios.post(`/auth`, { email, password })
  }

  static simpleTokenLogin({ email, password }: ILoginRequest): Promise<LoginResponse> {
    return axios.post(`/auth/login`, { email, password })
  }

packages/utils/functions/axios/createAxiosInstance/index.ts
```diff
@@ -55,7 +64,21 @@ export const createAxiosInstance = ({
...
+        tokenType === TokenTypes.jwt &&
+        error.response?.status === 401 &&
+        error.config.url !== REFRESH_TOKEN_URL
+      ) {
+        try {
+          const originalRequest = error.config
+          // eslint-disable-next-line @typescript-eslint/no-use-before-define
+          return await refreshAccessToken(cookieName, refreshCookieName, originalRequest)
+        } catch (refreshError) {
+          return Promise.reject(refreshError)
```
deboracosilveira: call endpoint to logout user and test if redirects user to login page

packages/authentication/package.json
```diff
@@ -12,13 +12,13 @@
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist"
   },
   "dependencies": {
-    "@baseapp-frontend/provider": "*",
-    "@baseapp-frontend/utils": "*",
```
deboracosilveira: revert

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@baseapp-frontend/core",
-  "version": "2.7.1",
+  "version": "2.7.2",
```
deboracosilveira: revert

deboracosilveira (approved):

packages/utils/functions/axios/createAxiosInstance/index.ts
pablogadhi: I have an implementation of this already, but I just realized that we would need to add nextjs as an implementation of the auth package to use the router and I'm not sure if that's ok. Will discuss it with the rest of the team.

Janekk: btw Should we use some event / pub-sub mechanism to fire of an event from axios interceptor that can be subscribet to in the authentication package to then trigger calling a `logout()`?

Janekk (approved):

packages/utils/functions/axios/createAxiosInstance/index.ts
pablogadhi: Made it like that, it now triggers an event, but put the actual subscribing and redirecting into the FE templates, bec we don't have next as a dependency even in the auth package. Not sure if we want to keep it like that for now. Maybe we can have a different package for next stuff

pablogadhi: Here's the PR: https://bitbucket.org/silverlogic/baseapp-frontend-template/pull-requests/17

packages/authentication/modules/user/useUser/index.ts
```diff
@@ -1,16 +1,28 @@
...
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import Cookies from 'js-cookie'
 
 import UserApi, { USER_API_KEY } from '../../../services/user'
 import { IUser } from '../../../types/user'
-import { IUseUser } from './types'
+import { IUseSimpleTokenUser, IUseUser } from './types'
 
-const useUser = <TUser extends Partial<IUser>>({
+const useUser = ({ cookieName = ACCESS_COOKIE_NAME }: IUseUser = {}): IUser | null => {
```
anicioalexandre: Could we still use the generics here? That gets pretty useful when projects change the User model and we can use the correct interface on it.
```suggestion
const useUser = <TUser extends Partial<IUser>, IJwtTokenSpecifics> ({ cookieName = ACCESS_COOKIE_NAME }: IUseUser = {}) => {
```

packages/authentication/modules/user/useUser/index.ts
```diff
@@ -1,16 +1,28 @@
...
+    try {
+      return decodeJWT<IUser>(token)
+    } catch (error) {
+      return null
+    }
+  }
+  return null
+}
+
+export const useSimpleTokenUser = <TUser extends Partial<IUser>>({
```
anicioalexandre: just a suggestion, I know we have no time for that. But I think would be cool to move this `useSimpleTokenUser` into its own file `user/useSimpleTokenUser`, maybe we add it as a TODO?

packages/authentication/package.json
```diff
@@ -18,7 +18,7 @@
     "@tanstack/react-query": "^4.29.12",
     "js-cookie": "^3.0.5",
     "react": "^18.2.0",
-    "react-hook-form": "^7.34.2",
+    "react-hook-form": "^7.46.1",
```
anicioalexandre, line 21: we might need a little story to bump the other packages that use the old versions too + the template

packages/utils/functions/axios/createAxiosInstance/index.ts
```diff
@@ -75,3 +103,34 @@ export const {
   requestInterceptorId: requestInterceptorIdForFiles,
   responseInterceptorId: responseInterceptorIdForFiles,
 } = createAxiosInstance({ file: true })
+
+export const refreshAccessToken = async (
```
anicioalexandre, line 108: this little guy might fit better at `utils/functions/token/refreshAccessToken`, maybe we can add a TODO to move it later?

anicioalexandre (approved):

anicioalexandre:
nice job @pablogadhi!

packages/authentication/modules/user/useUser/index.ts
```diff
@@ -1,16 +1,28 @@
...
 import { IUser } from '../../../types/user'
-import { IUseUser } from './types'
+import { IUseSimpleTokenUser, IUseUser } from './types'
 
-const useUser = <TUser extends Partial<IUser>>({
+const useUser = ({ cookieName = ACCESS_COOKIE_NAME }: IUseUser = {}): IUser | null => {
+  const token = Cookies.get(cookieName)
+  if (token) {
+    try {
+      return decodeJWT<IUser>(token)
```
anicioalexandre:
```suggestion
      return decodeJWT<TUser>(token)
```

packages/authentication/package.json
pablogadhi: Yeah, I was having an issue with the version that we had and updating solved it for me, but we probably want to do that for all of the packages if they start showing that issue as well.

packages/authentication/services/auth.ts
pablogadhi: I agree, I'll see if I can update the templates quickly and make all the endpoints in the auth package be like that

packages/utils/functions/axios/createAxiosInstance/index.ts
pablogadhi: The problem with that is that the function gets called inside the `createAxiosInstance` but at the same time the function needs access to the axios instance to call the refresh endpoint, so we would need to figure out that. I'll add the TODO tho so one day we handle it better.

packages/authentication/services/auth.ts
pablogadhi: Did it only for the `/login` one, since the other ones are for account management and should be under `/account` or something like that, but that's outside the scope of this story.

packages/authentication/modules/user/useUser/index.ts
pablogadhi: It was simple enough so I moved the hook to it's own folder together with the types and the tests

packages/authentication/modules/user/useUser/index.ts
pablogadhi: Implemented!

PR #61, opened 2023-10-04
GGL-160: replaces yup with zod

anicioalexandre:
This PR replaces `yup` with `zod` as a validation schema library. Some reasons that guided this decision:
* `zod` has better performance and a smaller bundle size compared with `yup`.
* `zod` has better type inferences and works better with Typescript in general, when compared with `yup`.
* both libraries have similar interfaces, but `zod` ends up having better in-built validations (like the `.email()` one).

Changes:
* authentication package update - `v1.20`
  - Replaces `yup` for `zod` as schema validation utility.
  - Add tests to `defaultValues` and `validationSchema` props to ensure that customization works.

* utils package update - `v1.3.1`
  - Add `ZOD_MESSAGE` constant with reusable validation messages.


Ronan-Fernandes (approved):

pt-tsl (approved):

packages/authentication/modules/access/useSignUp/index.ts
```diff
@@ -37,6 +37,7 @@ const useSignUp = <TRegisterRequest extends IRegisterRequest, TRegisterResponse
     try {
       await mutation.mutateAsync(values)
     } catch (error) {
+      console.log(error, 'error')
```
deboracosilveira, line 40: remove or update?

deboracosilveira (approved):

PR #72, opened 2023-12-05
BA-1232: pre-authenticated link & shorten links

patsluth-tsl:
Acceptance Criteria:

It should be able to generate pre-authenticated links for a user
** different pre authenticated links should be able to work, even if it's generated for the same user
** pre-auth links should work with JWT auth
** the token will still be valid as long as the underlying access token lives (base it on a separate Access Token that is set to ** expire in longer time than the normal access token, e.g. 1 week)
** if using Simple Auth (not JWT), support the expiration of pre-auth links too

Add the ability to generate shortened URLs links
** shorten by default

Inspiration (single use token):
** [https://bitbucket.org/silverlogic/ggl-web/src/ggl-post-mvp-epic/app/c/[shortCode]/](https://bitbucket.org/silverlogic/ggl-web/src/ggl-post-mvp-epic/app/c/%5BshortCode%5D/)
** [https://bitbucket.org/silverlogic/ggl-web/src/ggl-post-mvp-epic/app/pre-auth/[userId]/page.tsx](https://bitbucket.org/silverlogic/ggl-web/src/ggl-post-mvp-epic/app/pre-auth/%5BuserId%5D/page.tsx)
** https://bitbucket.org/silverlogic/ggl-api/pull-requests/2

add tests for both FE and BE implementations

Demo Link: TODO

### DIFF
--- a/packages/authentication/modules/access/usePreAuth/__tests__/usePreAuth.test.tsx
+++ b/packages/authentication/modules/access/usePreAuth/__tests__/usePreAuth.test.tsx
@@ -1,0 +1,51 @@
+import { ComponentWithProviders, MockAdapter, renderHook, waitFor } from '@baseapp-frontend/test'
+import { axios } from '@baseapp-frontend/utils'
+import { TokenTypes } from '@baseapp-frontend/utils/constants/token'
+
+import usePreAuth from '../index'
+
+// @ts-ignore TODO: (BA-1081) investigate AxiosRequestHeaders error
+export const axiosMock = new MockAdapter(axios)
+
+describe('usePreAuth jwt', () => {
+  test('should run onSuccess', async () => {
+    axiosMock.onPost('/auth/pre-auth/jwt').reply(200, {
+      refresh: 'fake-refresh-token',
+      access: 'fake-access-token',
+    })
+
+    const { result } = renderHook(
+      () =>
+        usePreAuth({
+          token: 'fake-pre-auth-token',
+          tokenType: TokenTypes.jwt,
+        }),
+      {
+        wrapper: ComponentWithProviders,
+      },
+    )
+
+    await waitFor(() => expect(result.current.query.isSuccess).toBe(true))
+  })
+})
+
+describe('usePreAuth auth_token', () => {
+  test('should run onSuccess', async () => {
+    axiosMock.onPost('/auth/pre-auth/auth-token').reply(200, {
+      token: 'fake-auth-token',
+    })
+
+    const { result } = renderHook(
+      () =>
+        usePreAuth({
+          token: 'fake-pre-auth-token',
+          tokenType: TokenTypes.simple,
+        }),
+      {
+        wrapper: ComponentWithProviders,
+      },
+    )
+
+    await waitFor(() => expect(result.current.query.isSuccess).toBe(true))
+  })
+})
--- a/packages/authentication/modules/access/usePreAuth/index.ts
+++ b/packages/authentication/modules/access/usePreAuth/index.ts
@@ -1,0 +1,58 @@
+import { ACCESS_COOKIE_NAME, REFRESH_COOKIE_NAME, TokenTypes } from '@baseapp-frontend/utils'
+
+import { useQuery, useQueryClient } from '@tanstack/react-query'
+import Cookies from 'js-cookie'
+
+import AuthApi, { PRE_AUTH_API_KEY } from '../../../services/auth'
+import { USER_API_KEY } from '../../../services/user'
+import { isJWTResponse } from '../../../utils/login'
+import { useSimpleTokenUser } from '../../user'
+import { IUsePreAuth } from './types'
+
+const usePreAuth = ({
+  token,
+  queryOptions = {},
+  tokenType = TokenTypes.jwt,
+  cookieName = ACCESS_COOKIE_NAME,
+  refreshCookieName = REFRESH_COOKIE_NAME,
+  ApiClass = AuthApi,
+}: IUsePreAuth) => {
+  const queryClient = useQueryClient()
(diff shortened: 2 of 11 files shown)


packages/authentication/modules/access/usePreAuth/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { TokenTypes } from '@baseapp-frontend/utils'
+
+import { ILoginJWTResponse, ILoginSimpleTokenResponse } from '../../../types/auth'
+
+export const isJWTResponse = (
```
anicioalexandre: I think we already have that function under `packages/authentication/modules/access/useLogin/utils.ts`, maybe we could move this file to `packages/authentication/utils/login.ts` and reuse in both places?

packages/authentication/modules/access/usePreAuth/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  cookieName = ACCESS_COOKIE_NAME,
+  refreshCookieName = REFRESH_COOKIE_NAME,
+  ApiClass = AuthApi,
+}: IUsePreAuth) => {
+  const queryClient = useQueryClient()
+  const { refetch: refetchUser } = useSimpleTokenUser({ options: { enabled: false } })
+
+  const query = useQuery({
+    staleTime: Infinity, // makes cache never expire automatically
+    enabled: !!token,
```
anicioalexandre, line 24: do we need that? Normally this endpoint doesn't require/allow a token. I think that would stop people who don't have any token set to run that query, no?

packages/authentication/modules/access/usePreAuth/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  tokenType = TokenTypes.jwt,
+  cookieName = ACCESS_COOKIE_NAME,
+  refreshCookieName = REFRESH_COOKIE_NAME,
+  ApiClass = AuthApi,
+}: IUsePreAuth) => {
+  const queryClient = useQueryClient()
+  const { refetch: refetchUser } = useSimpleTokenUser({ options: { enabled: false } })
+
+  const query = useQuery({
+    staleTime: Infinity, // makes cache never expire automatically
```
anicioalexandre: don't think we need that for this query, do we?

packages/authentication/modules/access/usePreAuth/index.ts
```diff
@@ -0,0 +1,60 @@
...
+          secure: process.env.NODE_ENV === 'production',
+        })
+      } else {
+        Cookies.set(cookieName, response.token, {
+          secure: process.env.NODE_ENV === 'production',
+        })
+      }
+      // by invalidating the cache we force a reload of /v1/users/me and the state used by useUser hook
+      queryClient.invalidateQueries(USER_API_KEY.getUser())
+      refetchUser()
```
anicioalexandre, line 44: let's use `resetQueries` instead of invalidate + refetching as we do [here](https://github.com/silverlogic/baseapp-frontend/blob/master/packages/authentication/modules/access/useLogout/index.ts#L20)
we had an issue during the logout that by just invalidating it wouldn't clear the previous cache.
so I think just `resetQueries` would be enough here, and then whenever we enter an authenticated page (that requires a user to be fetched) it will be fetched by a [withAuthentication HOC](https://bitbucket.org/silverlogic/baseapp-frontend-template/src/6698a3a054281305b12f7c0af17da33c0952b29c/.app-templates/mui-project/modules/auth/withAuthentication/index.tsx#lines-37) or something similar

packages/authentication/modules/access/usePreAuth/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  const queryClient = useQueryClient()
+  const { refetch: refetchUser } = useSimpleTokenUser({ options: { enabled: false } })
+
+  const query = useQuery({
+    staleTime: Infinity, // makes cache never expire automatically
+    enabled: !!token,
+    useErrorBoundary: false,
+    retry: false,
+    ...queryOptions, // needs to be placed bellow all overridable options
+    queryKey: ['preAuth', token, tokenType],
```
anicioalexandre: can we set something similar to this in the `packages/authentication/services/auth.ts`?
```
export const AUTH_API_KEY = {
  default: ['auth'],
  preAuth: (token: string, tokenType: TokenTypes) => [...AUTH_API_KEY.default, 'preAuth', token, tokenType] as const,
}
```
and use it here instead? just so we can reuse that same query key builder in other places if needed

.changeset/slow-waves-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@baseapp-frontend/authentication": patch
+---
+
+BA-1232-ba-fe-baseapp-pre-authenticated-link-shorten-links
```
anicioalexandre: just a suggestion, we usually do a `yarn version-packages` before committing because it's easier to edit the changelog and alter description if needed

packages/authentication/modules/access/usePreAuth/__tests__/usePreAuth.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+          token: 'fake-pre-auth-token',
+          tokenType: TokenTypes.simple,
+        }),
+      {
+        wrapper: ComponentWithProviders,
+      },
+    )
+
+    await waitFor(() => expect(result.current.query.isSuccess).toBe(true))
+  })
```
anicioalexandre, line 50: would be cool to have a test case where we set a cookie `123`, run the pre-auth hook, and get a `456` token after mocking the pre-auth API.
but I think that's too much for the points we have on that story, maybe we do that under a bug someday

anicioalexandre (changes requested):

anicioalexandre (approved):

PR #84, opened 2024-01-27
Fix refresh token flow for multipart requests

pablogadhi:
We don't handle the JWT refresh token flow with the axios instance meant for file uploads. This PR fixes that. We are changing the flow slightly to call the refresh endpoint before each request if we found out that the access token is expired because the formData gets consumed in the request and I couldn't find a nice way to keep a copy of it for the second request. This ends up having a little better performance since we make only 2 requests now, instead of 3, if the token expired. HOWEVER, there's this tiny tiny chance that the token expires in the middle of reaching the BE. If someone is concerned with that we can add a threshold in the token verification function to always refresh the token a little bit before it expires.

### DIFF
--- a/packages/utils/functions/token/getToken/index.ts
+++ b/packages/utils/functions/token/getToken/index.ts
@@ -1,10 +1,10 @@
 import ClientCookies from 'js-cookie'
-import { cookies as serverCookies } from 'next/headers'
 
 import { ACCESS_COOKIE_NAME } from '../../../constants/cookie'
 
-export const getToken = (cookieName = ACCESS_COOKIE_NAME) => {
+export const getToken = async (cookieName = ACCESS_COOKIE_NAME) => {
   if (typeof window === typeof undefined) {
+    const { cookies: serverCookies } = await import('next/headers')
     return serverCookies().get(cookieName)?.value
   }
   return ClientCookies.get(cookieName)
--- a/packages/utils/functions/token/refreshAccessToken/index.ts
+++ b/packages/utils/functions/token/refreshAccessToken/index.ts
@@ -1,0 +1,38 @@
+import _axios from 'axios'
+import Cookies from 'js-cookie'
+
+import { IJWTResponse } from '../../../types/jwt'
+
+const REFRESH_TOKEN_URL = '/auth/refresh'
+
+// We create an isolated axios instance to skip the interceptors
+export const simpleAxios = _axios.create({
+  baseURL: process.env.NEXT_PUBLIC_API_BASE_URL,
+})
+
+export const refreshAccessToken = async (cookieName: string, refreshCookieName: string) => {
+  const refreshToken = Cookies.get(refreshCookieName)
+
+  if (!refreshToken) {
+    return Promise.reject(new Error('No refresh token'))
+  }
+
+  try {
+    const responseData = (
+      await simpleAxios.post(REFRESH_TOKEN_URL, {
+        refresh: refreshToken,
+      })
+    ).data as IJWTResponse
+
+    Cookies.set(cookieName, responseData.access, {
+      secure: process.env.NODE_ENV === 'production',
+    })
+
+    return responseData.access
+  } catch (error) {
+    Cookies.remove(cookieName)
+    Cookies.remove(refreshCookieName)
+
+    return Promise.reject(error)
+  }
+}
--- a/packages/utils/hooks/useDebounce/index.ts
+++ b/packages/utils/hooks/useDebounce/index.ts
@@ -1,3 +1,5 @@
+'use client'
+
 import { useEffect, useRef } from 'react'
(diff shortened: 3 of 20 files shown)


packages/utils/hooks/useDebounce/index.ts
```diff
@@ -1,3 +1,5 @@
+'use client'
```
pablogadhi, line 1: I was having issues with an import of a utils function in a server side component because of these guys that use useState or useEffect, so I added the client directive.

pablogadhi:
I'll fix the tests on Monday

packages/utils/functions/token/getToken/index.ts
```diff
@@ -1,10 +1,10 @@
 import ClientCookies from 'js-cookie'
-import { cookies as serverCookies } from 'next/headers'
```
pablogadhi, line 2: This function was giving me some issues so I opted out from using it in the project that needed it. It looks like nextjs doesn't like us to use things from next/headers outside the server side. Maybe getting the token in the server vs the client should be separate functions in their own files. Don't know if using a dynamic import fixes it, I didn't test it out...

packages/utils/hooks/useDebounce/index.ts
Janekk: you mean while importing them on the template project?

Janekk (approved):

packages/utils/functions/token/refreshAccessToken/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const REFRESH_TOKEN_URL = '/auth/refresh'
+
+// We create an isolated axios instance to skip the interceptors
+const simpleAxios = _axios.create({
+  baseURL: process.env.NEXT_PUBLIC_API_BASE_URL,
+})
+
+export const refreshAccessToken = async (cookieName: string, refreshCookieName: string) => {
+  const refreshToken = Cookies.get(refreshCookieName)
```
anicioalexandre, line 14: will that work server side? maybe we can use the `getToken` instead?

pablogadhi: I was thinking that this function would only be meant to be used in the client side @anicioalexandre, because you can't use axios in the server side either and need to use nextjs's fetch. We are experimenting with a server-side version of it in a FAVIO project right now: https://bitbucket.org/silverlogic/hcmu-web/pull-requests/38#Lservices/fetch.tsT6, but we want to test it out for a couple of weeks before adding it to baseapp

packages/utils/hooks/useDebounce/index.ts
pablogadhi: Yes, importing them on the template project in server components. Because we do this barrel export thing in the main index.ts of the package @Janekk , and if you don't specify the exact absolute path, nextjs ends up trying to import client things on the server side.

packages/utils/functions/token/refreshAccessToken/index.ts
anicioalexandre: true! That fetch implementation is definitely something we should bring to the BA, let's test it then!

anicioalexandre (approved):

.changeset/big-actors-tap.md
```diff
@@ -0,0 +1,6 @@
+---
+"@baseapp-frontend/authentication": major
+"@baseapp-frontend/utils": major
```
anicioalexandre: I agree with the major bump, it would be nice if we could have a separate changelog for each package so we can guide people on what she/he should pay attention to when upgrading from a previous version

pablogadhi: Ok yeah I agree, I'll make sure that we have a changelog per package

packages/authentication/modules/user/useSimpleTokenUser/index.ts
```diff
@@ -5,13 +5,13 @@ import Cookies from 'js-cookie'
...
 import { IUser } from '../../../types/user'
-import { IUseSimpleTokenUser } from './types'
+import { IUseSimpleTokenUser, IUseSimpleTokenUserResult } from './types'
 
 const useSimpleTokenUser = <TUser extends Partial<IUser>>({
   options,
   cookieName = ACCESS_COOKIE_NAME,
   ApiClass = UserApi,
-}: IUseSimpleTokenUser<TUser> = {}) => {
+}: IUseSimpleTokenUser<TUser> = {}): IUseSimpleTokenUserResult<TUser> => {
```
anicioalexandre: do we need that? letting TS infer the return type is often easier and more accurate

pablogadhi: It was just for a test so I moved the type to the test and let this one be inferred

PR #87, opened 2024-02-27
Add multi-param support to useDebounce, add useDebouncedValue

nick-prat:
Updated the `useDebounce` to allow for more than one parameter. 

useDebouncedValue lets you debounce a value instead of function, something like
```
const [searchValue, setSearchValue] = useState("")
const debouncedSearchValue = useDebouncedValue(searchValue, 150)
```
this lets the UI update immidietly be tracking `searchValue` but the `debouncedSearchValue` only updated after 150ms (and only once per 150ms period)

@deboracosilveira I believe you originally wrote the `useDebouncedValue` for WAYS :smile:

### DIFF
--- a/packages/utils/hooks/useDebounce/index.ts
+++ b/packages/utils/hooks/useDebounce/index.ts
@@ -1,11 +1,11 @@
 'use client'
 
-import { useEffect, useRef } from 'react'
+import { useEffect, useState, useRef } from 'react'
 
 import debounce from 'lodash/debounce'
 
-const useDebounce = <T extends (params: any) => any>(
-  functionToDebounce: T,
+const useDebounce = <T extends any[], U>(
+  functionToDebounce: (...params: T) => U,
   { debounceTime = 400 } = {},
 ) => {
   const debouncedFuncRef = useRef(debounce(functionToDebounce, debounceTime, { trailing: true }))
@@ -20,4 +20,21 @@ const useDebounce = <T extends (params: any) => any>(
   return { debouncedFunction: debouncedFuncRef.current }
 }
 
-export default useDebounce
+const useDebouncedValue = <T>(value: T, debounceTime: number) => {
+  const [debouncedValue, setDebouncedValue] = useState(value)
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedValue(value)
+    }, debounceTime)
+
+    return () => {
+      clearTimeout(handler)
+    }
+  }, [value, debounceTime])
+
+  return debouncedValue
+}
+
+
+export {useDebounce as default, useDebouncedValue}


packages/utils/hooks/useDebounce/index.ts
```diff
@@ -20,4 +20,21 @@ const useDebounce = <T extends (params: any) => any>(
...
+    return () => {
+      clearTimeout(handler)
+    }
+  }, [value, debounceTime])
+
+  return debouncedValue
+}
+
+
+export {useDebounce as default, useDebouncedValue}
```
nick-prat, line 40: I've exported as default here, as it's the current behaviour.

packages/utils/hooks/useDebounce/index.ts
```diff
@@ -20,4 +20,21 @@ const useDebounce = <T extends (params: any) => any>(
   return { debouncedFunction: debouncedFuncRef.current }
```
nick-prat, line 20: I also wanted to ask why we return this as an object instead of just the value?

packages/utils/hooks/useDebounce/index.ts
```diff
@@ -1,11 +1,11 @@
...
 
-import { useEffect, useRef } from 'react'
+import { useEffect, useState, useRef } from 'react'
 
 import debounce from 'lodash/debounce'
 
-const useDebounce = <T extends (params: any) => any>(
-  functionToDebounce: T,
+const useDebounce = <T extends any[], U>(
+  functionToDebounce: (...params: T) => U,
```
anicioalexandre, line 8: I don't understand why we would need to manually define a return type here, could you explain?

packages/utils/hooks/useDebounce/index.ts
anicioalexandre: I don't think there was a real reason for that at the time, I guess because we were thinking about returning more data. I'm ok to return only the function instead

packages/utils/hooks/useDebounce/index.ts
```diff
@@ -20,4 +20,21 @@ const useDebounce = <T extends (params: any) => any>(
   return { debouncedFunction: debouncedFuncRef.current }
 }
 
-export default useDebounce
+const useDebouncedValue = <T>(value: T, debounceTime: number) => {
```
anicioalexandre, line 23: maybe we move this to its own file?

packages/utils/hooks/useDebounce/index.ts
nick-prat: I guess this could have been 
```typescript
const useDebounce = <T extends (...params: any[]) => any>(
  functionToDebounce: T,
  { debounceTime = 400 } = {},
) => ...
```
I just added it as it's how I'd personally write this generic, but I think it's the same thing functionally.

packages/utils/hooks/useDebounce/index.ts
nick-prat: Yup I can do that :+1:

priscilladeroode:
Is this PR abandoned? Does it still makes sense @nick-prat and @anicioalexandre? If not can we close this PR?

PR #101, opened 2024-06-27
BA-1454: setup internationalization

priscilladeroode:
- Make sure all HTTP calls, including GraphQL ones if possible, sends the Accept-Language HEADER based on user preferences

- Store the user preference on cookies and/or on user model


packages/provider/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@baseapp-frontend/provider",
   "description": "Providers for React Query and Emotion.",
-  "version": "2.0.1",
+  "version": "3.0.0",
```
anicioalexandre: looks like it should've been a patch update instead

packages/graphql/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@baseapp-frontend/graphql",
   "description": "GraphQL configurations and utilities",
-  "version": "1.1.2",
+  "version": "2.0.0",
```
anicioalexandre: looks like it should've been a patch update instead

packages/design-system-mui/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@baseapp-frontend/design-system-mui",
   "description": "Design System components and configurations.",
-  "version": "2.0.0",
+  "version": "3.0.0",
```
anicioalexandre: looks like it should've been a patch update instead

packages/authentication/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@baseapp-frontend/authentication",
   "description": "Authentication modules.",
-  "version": "3.1.0",
+  "version": "4.0.0",
```
anicioalexandre: looks like it should've been a minor update instead

packages/authentication/modules/user/useJWTUser/index.ts
```diff
@@ -50,6 +51,12 @@ const useJWTUser = <TUser extends Partial<User> & JWTContent>({
     }
   }, [query.error])
 
+  useEffect(() => {
```
anicioalexandre: I would suggest setting this cooking on the template' middleware ([around here](https://bitbucket.org/silverlogic/baseapp-frontend-template/src/88f345bbeba473356d13435dd05caa111259dfd7/middleware.ts#lines-37)) just to make sure we're always setting this cookie (not relying on `useJWTUser` or similar)

but if feels to me that a language cookie would not be necessary since we have the data available on the authorization cookie already (after decoding it). I think that we could alter `getLanguage` a little bit
```
export const getLanguage = (
  cookieName = ACCESS_COOKIE_NAME,
  { noSSR = false }: ServerSideRenderingOption = {},
) => {
  const user = getUser({ noSSR })
  if (!(user?.preferredLanguage in LanguagesEnum)) return null

  return user?.preferredLanguage
}

```
then on the fetch instance:
```
...
accessCookieName,
...
  const language = getLanguage(accessCookieName, { noSSR: false })
  if (language) {
    fetchOptions.headers!['Accept-Language'] = language
  }

```

packages/provider/package.json
priscilladeroode: Updated.

packages/authentication/modules/user/useJWTUser/index.ts
priscilladeroode: Hey @anicioalexandre implemented on the middleware :)

anicioalexandre (approved):

anicioalexandre (approved):

PR #102, opened 2024-08-07
BA-1486: modularize navigation

deboracosilveira:
*Context*
This user story aims to modularize the Navigation feature, including the header and sidebar, by moving them to the `components` package under the @baseapp-frontend organization. The goal is to ensure all navigation variations remain available and allow users to customize their navigation data easily.

*Given* the need to modularize the Navigation feature (header & sidebar).
*When* the features are moved to a new package under the `@baseapp-frontend` organization
*Then* the features should support the following customizations:
* all navigation variations should remain available (centered, vertical, mini, and horizontal)
* users can easily input their own navigation data (path routes, icons, sub-routes, etc...), ensuring type safety with TypeScript
* since `useUISettings` will remain on the template, the components that use it (e.g. `Header` and `NavToggleButton`) should get `settings` or `setSettings` via props

*General Requirements*
* the new package should be easily integrated with existing projects
* documentation should be provided to guide users on how to implement and customize these components
* update the BaseApp Frontend Template to use the new components and remove existing code related to these features
* include Storybook documentation (use comments storybook as an inspiration)


packages/components/modules/navigations/Header/__storybook__/ProjectLogoCondensed.tsx
anicioalexandre, line 1: let's move this to `packages/components/.storybook/components/ProjectLogoCondensed` since it's being reused by multiple storybooks

packages/components/modules/navigations/Header/__storybook__/navigationMocks.tsx
anicioalexandre, line 1: this could also be moved to `packages/components/.storybook/__mocks__/navigation.ts` since it's being reused by multiple storybooks

packages/components/modules/navigations/Header/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        )}
+        <IconButton
+          onClick={onOpenNav}
+          sx={{
+            display: { xs: 'flex', lg: 'none' },
+          }}
+        >
+          <MenuIcon />
+        </IconButton>
+        <AccountMenu additionalComponent={additionalComponent} {...AccountMenuProps}>
```
anicioalexandre: I think we could let `additionalComponent` be defined in `AccountMenuProps` only. But I understand that by separating it makes it easier when rendering  `NavigationLayout` on the template

packages/components/modules/navigations/NavigationLayout/__storybook__/stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+}
+
+export const CustomAccountMenu: Story = {
+  parameters: {
+    userType: 'valid',
+  },
+  args: {
+    ...ValidUser.args,
+    AccountMenu: () => <div>Custom AccountMenu</div>,
```
anicioalexandre: maybe we could apply `width: 100%; justify-content: center; display: flex;` just so the project icon can stay on the left side

packages/components/.storybook/preview.tsx
```diff
@@ -1,62 +1,14 @@
...
-        >
-          <NotificationProvider>
-            <Story />
-          </NotificationProvider>
-        </ThemeProvider>
-      </React.Suspense>
-    </RelayTestProvider>
-  )
-}
-
```
anicioalexandre, line 59: let's add a `layout: 'fullscreen'` I think it's better for visualizing the headers (it removes the body's padding). I'm wondering if we should do it for each instead of doing it globally though... because the padding is nice for the social stories
| without padding  | with padding |
| ------------- | ------------- |
| Screenshot from 2024-08-13 20-51-50 | Screenshot from 2024-08-13 20-52-03 |

```
const preview: Preview = {
  decorators: [withProviders],
  loaders: [mswLoader],
  parameters: {
    layout: 'fullscreen',
    ...
```

packages/components/.storybook/preview.tsx
anicioalexandre, line 1: we could set an order under `storySort` so we prioritize the most important stories
```
      storySort: (a, b) => {
        const order = [
          // Navigation
          'NavigationLayout',
          'Header',
          'AccountPopover',
          "NavMini",
          "NavHorizontal",
          "NavCentered",
          "NavVertical",
          // Social
          'Comments',
          'CommentsList',
          'CommentCreate',
          'CommentUpdate',
          'CommentItem',
          'CommentUpsertActions',
          'ReactionButton',
          'Timestamp',
        ]
```

packages/design-system/components/HeaderShadow/index.tsx
anicioalexandre, line 1: this one might never be reused I think, maybe move it closer to `packages/components/modules/navigations/NavHorizontal`?

anicioalexandre (commented):
nice work!

packages/components/.storybook/preview.tsx
```diff
@@ -1,62 +1,14 @@
...
-} from '@baseapp-frontend/design-system'
-import { RelayTestProvider } from '@baseapp-frontend/graphql'
-import { NotificationProvider } from '@baseapp-frontend/utils'
-
-import type { Preview, StoryContext, StoryFn } from '@storybook/react'
+import type { Preview } from '@storybook/react'
 import { initialize, mswLoader } from 'msw-storybook-addon'
 
 import '../styles/tailwind/globals.css'
+import { handlers } from './__mocks__/handlers'
```
anicioalexandre, line 5: we could add this css file:

```
// packages/components/.storybook/overrides.css
.simplebar-wrapper {
    height: 64px;
}
```
and import it on the preview file to fix the nav horizontal bug
```
import './overrides.css'
```

packages/components/.storybook/preview.tsx
deboracosilveira: added on the specific ones where it makes a difference

anicioalexandre (approved):

packages/utils/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @baseapp-frontend/utils
 
+## 2.5.3
```
anicioalexandre: it should be `2.5.4`

PR #103, opened 2024-08-13
BA-1516: create env object

anicioalexandre:
* `utils` package update - `v 2.5.2`
  - Create and use `env` object to access `process.env` variables.

* `authentication` package update - `v 3.2.2`
  - Use `env` object to access `process.env` variables.
  - Updated dependencies
    - @baseapp-frontend/utils@2.5.2

* `graphql` package update - `v 1.1.5`
  - Use `env` object to access `process.env` variables.
  - Updated dependencies
    - @baseapp-frontend/utils@2.5.2


Janekk (approved):

packages/utils/functions/env/index.ts
nossila, line 1: How would we handle a new/specific env variable in the context of a project?

rafaelhdr (approved):

packages/utils/functions/env/index.ts
anicioalexandre: This one is to be used inside the packages only. On projects we're meant to use this [one](https://bitbucket.org/silverlogic/baseapp-frontend-template/pull-requests/82/diff#chg_utils/env.ts_newline1).  I'll change the name to `templateEnv` so it is less confusing

PR #108, opened 2024-08-22
BA-1533: Added useLogoOverrides hook and Dropzone component to Design…

junioresc:
… System, Added getImageString to Utils pkg

Acceptance Criteria 
Context
This user story focuses on making it easy to change the BaseApp logo to a prospect client's logo for quick demos.

Given the need to change logos on demand.
When the feature is implemented.
Then it should support the following functionalities:

Dynamic Logo Display:

create/use the Theme section inside the user's settings page and build the Logo section there

create image inputs for normal logo and condensed logos

the inputs show the current logo as the preview image

it should accept only SVG formats

after the image upload and saving, we should convert the SVG to base64 and store it in the local storage

ensure all static logo images use the local storage base64 image as the first option; if it's not found, use the static image

the logo change is meant to be temporary, meaning that if the person didn't upload any image, the app shall render the static image instead

the change should be only applicable to the user session (e.g., someone accessing the site from a different computer should not be able to detect the logo change)

only superusers will have permissions to change any logo image 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp-Minimal-v5.4.0?node-id=2085-40113&t=zxCMuRHZ1Rp3OOt5-0 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/31638

### DIFF
--- a/packages/design-system/hooks/useLogoOverrides/index.ts
+++ b/packages/design-system/hooks/useLogoOverrides/index.ts
@@ -1,0 +1,49 @@
+'use client'
+
+import { ServerSideRenderingOption, getCookie, setCookie } from '@baseapp-frontend/utils'
+
+import { atom, useAtom } from 'jotai'
+
+import { DEFAULT_LOGO_KEY, DEFAULT_LOGO_SETTINGS } from './constants'
+import { LogoOverrides } from './types'
+
+export const getLogoOverridesFromCookie = ({ noSSR = true }: ServerSideRenderingOption = {}) => {
+  const settings = getCookie<LogoOverrides>(DEFAULT_LOGO_KEY, { noSSR }) ?? DEFAULT_LOGO_SETTINGS
+
+  return settings
+}
+
+const initialLogos = getLogoOverridesFromCookie()
+
+const settingsAtom = atom<LogoOverrides>(initialLogos)
+
+/**
+ * By using `useLogoOverrides` with the `noSSR` option set to `false`, causes Next.js to dynamically render the affected pages, instead of statically rendering them.
+ */
+const useLogoOverrides = ({ noSSR = true }: ServerSideRenderingOption = {}) => {
+  const [settings, setSettings] = useAtom(settingsAtom)
+  const isSSR = typeof window === typeof undefined
+
+  const handleSetLogoOverrides = (newLogos: Partial<LogoOverrides>) => {
+    setSettings((prevLogos: LogoOverrides) => {
+      const updatedLogoOverrides = { ...prevLogos, ...newLogos }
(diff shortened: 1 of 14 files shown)


pablogadhi:
@junioresc The updated logo is not being persisted in the localstorage so it turns back to the initial one when you navigate to another page. According to the AC, it should be doing that, so maybe we need to update the hook to save the image in the localstorage instead of a jotai atom state. See loom: https://www.loom.com/share/f5b1523aefe64988a63f4bf500617057?sid=7c7b4868-e100-4bd2-bad2-e78bf43d9279

packages/design-system/hooks/useLogoOverrides/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const initialLogos = getLogoOverridesFromCookie()
+
+const settingsAtom = atom<LogoOverrides>(initialLogos)
+
+/**
+ * By using `useLogoOverrides` with the `noSSR` option set to `false`, causes Next.js to dynamically render the affected pages, instead of statically rendering them.
+ */
+const useLogoOverrides = ({ noSSR = true }: ServerSideRenderingOption = {}) => {
+  const [settings, setSettings] = useAtom(settingsAtom)
```
pablogadhi, line 24: You could use the atomWithStorage here to solve the persistence issue mentioned before, but I don't know how others feel about adding another state management library just for this. Personally I think it's fine, since jotai it's supposed to have a different use case than zustand.

pablogadhi (changes requested):

packages/design-system/hooks/useLogoOverrides/index.ts
anicioalexandre: I think we can go with a simpler solution (use local storage at the template level) since this feature was made to allow the sales team to demo BA with clients. As it should be built now, it's not something our regular clients will use. `jotai` is being used on the `useUISettings` because we needed to be able to optionally get the initial state from a cookie or not

anicioalexandre: In the `useUISettings` we used cookies to store the theme information so we can get it from the client and from the server (and by doing so, we avoid glitches from server to client). In this case, I wouldn't really bother using local storage (client-only) and have this small glitch, since it's a feature for demonstrations

pablogadhi: Ok, if we are fine with not persisting thins along pages, it's fine by me to not bother with the localstorage.

pablogadhi (approved):

PR #111, opened 2024-09-06
BA-1548: changed useLogoOverrides to use localStorage...

junioresc:
…instead of cookies, moved project logo to design-system.

- Previously useLogoOverrides was implemented using cookies but that has a size limit of 4kb, if a large SVG image was uploaded by the user then it would not persist upon navigating to another page. The logic was changed so that it uses localStorage to persist larger images.

- Created ProjectLogo component to completed task left by @anicioalexandre on [BA-1415: Change Logos on Demand
](https://bitbucket.org/silverlogic/baseapp-frontend-template/pull-requests/80/overview)
> create a Logo component under components/design-system and move the logic there. We should also replace the logo usage throughout the app (we use the BA default and condensed logos in many places (header, login, sigin up, forgot password….)

### DIFF
--- a/packages/design-system/CHANGELOG.md
+++ b/packages/design-system/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @baseapp-frontend/design-system
 
+## 0.0.8
+
+### Patch Changes
+
+- Changed useLogoOverrides to use localStorage instead of cookies, moved project logo to design-system
+
 ## 0.0.7
 
 ### Patch Changes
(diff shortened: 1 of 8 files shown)


pablogadhi (approved):

packages/design-system/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @baseapp-frontend/design-system
 
+## 0.0.8
+
+### Patch Changes
+
+- BA-1548: changed useLogoOverrides to use localStorage instead of cookies, moved project logo to design-system
```
deboracosilveira: i think we can remove the jira id from here

deboracosilveira (approved):

PR #113, opened 2024-09-10
BA: Modularize in app notifications

Ronan-Fernandes:
*Context*
This user story aims to modularize the In-App Notifications features by moving them to the `components` package under the @baseapp-frontend organization. The goal is to enhance customizability and ease of integration for different projects and facilitate updates on these features.

*Given* the need to modularize the In-App Notifications feature.
*When* the features are moved to a new package under the @baseapp-frontend organization.
*Then* the features should support the following customizations:

* developer can change the popover menu icon.
* developer can provide custom loading and empty states.
* developer can customize drawer props (size, animations, etc.).

*General Requirements*

* the new package should be easily integrated with existing projects.
* documentation should be provided to guide users on how to implement and customize these components.
* update the BaseApp Frontend Template to use the new components and remove existing code related to the features

* include Storybook documentation 

Release notes:
* `components` package update - `v 0.0.5`
   - Add Notifications Component
   - Updated dependencies
      - @baseapp-frontend/design-system@0.0.10
 
* `design-system` package update - `v 0.0.10`
  - Add NotificationBellIcon


packages/components/modules/notifications/NotificationsPopover/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }}
+        {...restDrawerProps}
+      >
+        <NotificationsList setIsDrawerOpened={setIsDrawerOpened} {...NotificationsListProps} />
+      </Drawer>
+    </>
+  )
+}
+
+export default NotificationsPopover
```
anicioalexandre: we need to wrap this component in a Suspense boundary so we can show the proper loading (instead of triggering the application suspense boundary, which in the case of baseapp-template is `app/loading.tsx`)

https://relay.dev/docs/api-reference/use-lazy-load-query/
https://relay.dev/docs/guided-tour/rendering/loading-states/

```
const NotificationsPopoverSuspended = () => (
  <Suspense
    fallback={
      <IconButton disabled>
        <NotificationBellIcon color="secondary" />
      </IconButton>
    }
  >
    <NotificationsPopover />
  </Suspense>
)

export default NotificationsPopoverSuspended
```

packages/components/modules/notifications/NotificationsList/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+            </IconButton>
+          )}
+        </Box>
+      </HeaderContainer>
+      {renderContent()}
+    </Box>
+  )
+}
+
+export default NotificationsList
```
anicioalexandre: we need to wrap this component in a Suspense boundary so we can show the proper loading (instead of triggering the application suspense boundary, which in the case of baseapp-template is `app/loading.tsx`)

https://relay.dev/docs/api-reference/use-lazy-load-query/
https://relay.dev/docs/guided-tour/rendering/loading-states/

```
const NotificationsListSuspended: FC<NotificationsListProps> = (props) => (
  <Suspense fallback={<LoadingState />}>
    <NotificationsList {...props} />
  </Suspense>
)

export default NotificationsListSuspended
```

Ronan-Fernandes: nice catch, I updated

packages/components/modules/notifications/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const NOTIFICATION_VERB = {
+  commentCreated: 'COMMENTS.COMMENT_CREATED',
```
anicioalexandre, line 2: why did we change that?

Ronan-Fernandes: This change is based on the changes of this PR https://github.com/silverlogic/baseapp-backend/pull/132

anicioalexandre (approved):

PR #119, opened 2024-10-10
BA-1636: multiple profiles profile switching

Hercilio1:
> As a user, on the BaseApp Profile,I would like to Create an alternate profile, In order to manage different activies or personas within my account.

**Original Story:**  
Acceptance Criteria 
Business Rules

Given the profile has been created, then it  should be displayed in the "Switch Profile List", where we list all user's profiles

The current selected profile should be highlighted and with a checkmark to make it clear that its the active profile being used

we need to store this selection in a global state and persist it during reloads (sidebar with AA)

The list should show a maximum of 5 profiles

After selecting another profile, reload the page in order to load the platform with the new profile.

Make the app know which profile is selected, so that the social interactions show the profile as the author of the interactions.

Change the Avatar to the selected profile's avatar.

***Validation Rules***

Given the user has succesfully switch to another profile, display a toast to inform that.

Copy Text
*Switched profile to [Profile Name] 

Current behavior 

Expected behavior 

Code Snippet 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3654-40031&node-type=frame&t=MBvMfObBLgXLqFQv-0 

Notes 
We might need another story to handle the comments and reactions to reflect that those where done by the profile. 

Approvd 
https://app.approvd.io/projects/BA/stories/34197 

---

**Setup:**
* Clone these repos in the same folder:
  * https://bitbucket.org/silverlogic/baseapp-frontend-template/src/master/
  * https://bitbucket.org/silverlogic/baseapp-backend-template/src/master/
  * https://github.com/silverlogic/baseapp-frontend
* Get the BE (baseapp-backend-template) running normally.
* Switch the branches of the FE projects:
  * baseapp-backend-template: git checkout BA-1659-fe-templates-multiple-profiles-profile-switching
  * baseapp-frontend: git checkout BA-1636-fe-multiple-profiles-profile-switching
* Replace the following deps of the FE template (baseapp-frontend-template) package.json:
  ```
  "@baseapp-frontend/components": "file:../baseapp-frontend/packages/components",
  "@baseapp-frontend/design-system": "file:../baseapp-frontend/packages/design-system",
  ```
* Run the following commands in the root of the FE template (baseapp-frontend-template):
  ```
  yarn install --force
  rm -rf .next
  yarn dev
  ```

**Demo:** https://www.loom.com/share/12bef9e4ae124e36916720646814b9cf

---

**Notes:** The cypress tests are not complete yet. I'll finish it, but I would like the PR to get reviewed already.


packages/components/modules/profiles/hooks/useCurrentProfile/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @returns {currentProfile, setCurrentProfile}
+ */
+const useCurrentProfile = () => {
+  const [currentProfile, setCurrentProfile] = useAtom(currentProfileAtom)
+
+  const isSSR = typeof window === 'undefined'
+  const shouldFetchProfile = !isSSR && !currentProfile?.profile
+
+  const { me } = useLazyLoadQuery<UserProfileQueryType>(
```
nossila: We can't use `useLazyLoadQuery` here, this will make all components wrap suspense, and we want to avoid this. `useCurrentProfile` should only return the current profile and a method to change, lets avoid any possible API call here.

ThePrestwood:
Will show more than 5 profiles
Screenshot 2024-10-14 at 1 24 15 PM

ThePrestwood:
When I log in using a different browser it uses the default profile opposed to my last selected profile.
https://www.loom.com/share/e18fc8bd2912471c8e418d86cea58cdb

ThePrestwood:
Profile names should validate
image

packages/components/modules/profiles/hooks/useCurrentProfile/index.tsx
Hercilio1: Done. I moved it to inside a provider and removed the useLazyLoadQuery

packages/components/modules/profiles/context/CurrentProfileProvider/types.ts
```diff
@@ -0,0 +1,12 @@
+import { ProfileItemFragment$data } from '../../../../__generated__/ProfileItemFragment.graphql'
+
+export type CurrentProfileState = {
+  profile?: ProfileItemFragment$data
+  userId?: number
+}
+
+type CurrentProfileFunctions = {
+  setCurrentProfile: (newCurrentProfiles: Partial<CurrentProfileState>) => void
```
deboracosilveira: remove plural newCurrentProfiles ?

packages/components/modules/profiles/ProfileAccountPopover/ProfileMenuItem/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  profileRef,
+  currentProfile,
+  onProfileChange,
+  avatarProps = {},
+  width = 36,
+  height = 36,
+}) => {
+  const profile = readInlineData<ProfileItemFragment$key>(ProfileItemFragment, profileRef)
+
+  const handle = profile.urlPath?.path
```
deboracosilveira: guess we could use a more intuitive name, like routePath

deboracosilveira:
I think we could simplify the components structure a bit. My suggestions are:
- Rename ProfileAccountSections to ProfilePopover, since is basically a diff version of AccountPopover
- Simplify ProfilesSubmenusList name to ProfileList since they are basically a list of menu items, not exactly submenus. Also update all names related to this.
- Make AccountPopover props names more clear, rn we cant tell the diff between accountSection, menuSection and extraSections just by reading

deboracosilveira (changes requested):

packages/components/modules/navigations/Header/AccountMenu/AccountAvatar/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { AccountAvatarProps } from './types'
+
+const AccountAvatar: FC<AccountAvatarProps> = ({ popoverOpen, popoverOnOpen, ...props }) => {
+  const [isMounted, setIsMounted] = useState(false)
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
+
+  if (!isMounted) {
```
anicioalexandre: on the server side, we want to keep leveraging the user JWT and display the user image. Although the avatar won't be clickable,  just rendering the ClickableAvatar provides a smoother transition between server and client, thus a better UX. Let's revert that change

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }))
+  })
+
+  it('should render the account popover and be able to interact with it', () => {
+    const { environment } = createTestEnvironment()
+
+    cy.mount(<AccountPopoverForTesting environment={environment} />)
+
+    cy.findByRole('button').click()
+    cy.get('.MuiPaper-root').should('exist')
```
anicioalexandre: Let's avoid using selectors that don’t prioritize accessibility. Instead, we should use Testing Library selectors like findByRole, findByPlaceholderText, and findByLabelText. These DOM-testing queries encourage better accessibility practices by focusing on the component's roles and labels, which makes our tests more meaningful and our components more accessible in the process. Please refer to https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/comments/Comments/__tests__/Comments.cy.tsx

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
```diff
@@ -0,0 +1,146 @@
...
+        onClick: cy.stub().as('menu2'),
+      },
+    ]
+    cy.mount(<AccountPopoverForTesting environment={environment} menuItems={menusMock} />)
+
+    cy.findByRole('button').click()
+
+    cy.contains('li', /menu1/i).click()
+    cy.get('@menu1').should('have.been.calledOnce')
+    cy.wait(300) // wait for the menu to close
```
anicioalexandre: I didn't get why we have that many waits on the test cases, are those really necessary?

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+    cy.findByRole('button').click()
+
+    cy.contains('div', /account section/i).should('exist')
+    cy.contains('div', /menu section/i).should('exist')
+    cy.contains('div', /account actions section/i).should('exist')
+  })
+
+  it('should hide custom section', () => {
```
anicioalexandre: We should aim to include as many relevant assertions as possible within a single test case to cover more scenarios efficiently instead of doing one test scenario for just a few test cases. It will avoid unnecessary mounts and will make the tests faster. We can use `cy.step` to organize as many scenarios as we do into a single test case. Refer to https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/comments/Comments/__tests__/Comments.cy.tsx for more examples

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    cy.findByRole('button').click()
+
+    cy.contains('div', /account section/i).should('not.exist')
+  })
+
+  it('should add and hide extra sections', () => {
+    const { environment } = createTestEnvironment()
+
+    const extraSectionControllersListMock = [
```
anicioalexandre: I didn't really get how we are mocking the tests, but mocking the graphql operation itself using `resolveMostRecentOperation` would probably be more effective

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,135 @@
...
 }) => {
   const { user } = useJWTUser<BaseUser & JWTContent>()
   const { logout } = useLogout()
+
   const popover = usePopover()
 
-  const handleLogout = async () => {
+  const handleOnClose = () => {
+    if (onCloseCallback) {
+      onCloseCallback()
```
anicioalexandre: can avoid the if statement by doing
```suggestion
      onCloseCallback?.()
```

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,135 @@
...
-    popover.onClose()
+    handleOnClose()
   }
 
+  const renderItems = (items: ComponentItems) =>
+    items.map((item, index) => {
+      if (!item) {
+        return null
+      }
+      if (typeof item === 'function') {
```
anicioalexandre: do you want to check if it's a react component? if so you can use `isValidElement`

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,135 @@
...
+        const Component = item as FC
+        return <Component key={index} /> // eslint-disable-line react/no-array-index-key
+      }
+      return <Fragment key={index}>{item}</Fragment> // eslint-disable-line react/no-array-index-key
```
anicioalexandre: when would that happen? like rendering some string?

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,135 @@
...
+        alt="User avatar"
+        popoverOpen={Boolean(popover.open)}
+        popoverOnOpen={popover.onOpen}
+      />
+
+      <Popover
+        open={popover.open}
+        onClose={handleOnClose}
+        sx={{ ...DefaultPopoverStyles, ...PopoverStyles }}
       >
```
anicioalexandre, line 82: I got the idea behind the controllers, but it looks over complex to me. We could move these sections into components and make them overridable by the template. I think we should do something like:

```
const AccountPopover: FC<AccountPopoverProps> = ({
  // ...
  CurrentUser = DefaultCurrentUser,
  SwitchProfileMenu = DefaultSwitchProfileMenu,
  MenuItems = DefaultMenuItems,
  LogoutItem = DefaultLogoutItem,
}) => {
  // ...
  return (
    <>
      <ClickableAvatar
        src={user?.avatar?.small}
        alt={user?.email}
        onClick={popover.onOpen}
        isOpen={!!popover.open}
      >
        {user?.firstName && user?.lastName ? `${user?.firstName[0]}${user?.lastName[0]}` : ''}
      </ClickableAvatar>
      <Popover
        open={popover.open}
        onClose={handleOnClose}
        sx={{ ...DefaultPopoverStyles, ...PopoverStyles }}
      >
        <CurrentProfile /> // why name it a CurrentUserPlaceholder ?
        <SwitchProfileMenu /> // make sure that's an actual component

        {!!CurrentUser && !!SwitchProfileMenu && <Divider sx={{ borderStyle: 'solid' }} />}
        
        <MenuItems menuItems={menuItems} /> // create this new component rendering the menu items, it should accept a menuItems prop

        {!!LogoutItem && <Divider sx={{ borderStyle: 'solid' }} />}

        <LogoutItem /> // move logout logic inside this component
      </Popover>
    </>
  )
}
```

it think that should provide already a decent customization option for a v1. If we wanna provide a couple more granular customization options, we could expose these component props, like:

```
const AccountPopover: FC<AccountPopoverProps> = ({
  // ...
  SwitchProfileMenu = DefaultSwitchProfileMenu,
  SwitchProfileMenuProps = {},
  LogoutItem = DefaultLogoutItem,
  LogoutItemProps = {}
}) => {
  // ...
  return (
    <>
      <Popover>
        // ...
        <SwitchProfileMenu {...SwitchProfileMenuProps} />

        <LogoutItem {...LogoutItemProps} />
      </Popover>
    </>
  )
}
```

then let's say on the template, we don't want any of these profile features, we could just do:
```
    <BaseNavigationLayout
       // ...
      AccountMenuProps={{
        // ...
        AccountPopoverProps: {
          CurrentProfile: null,
          SwitchProfileMenu: null
        }
      }}
    >
      {children}
    </BaseNavigationLayout>
```

packages/components/modules/profiles/ProfilePopover/ProfileMenuItem/index.tsx
anicioalexandre, line 1: this component should be under the ProfilesList directory since it's only being used by that component

packages/components/modules/navigations/Header/AccountMenu/CurrentUserPlaceholder/types.d.ts
anicioalexandre, line 1: let's always declare our types on .ts files, and never on declaration .dts files. Those are meant to hold declarations (not implementations) and public types

packages/components/modules/profiles/ProfilePopover/ProfileMenuItem/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const ProfileMenuItem: FC<ProfileMenuItemProps> = ({
+  profileRef,
+  currentProfile,
+  onProfileChange,
+  avatarProps = {},
+  width = 36,
+  height = 36,
+}) => {
+  const profile = readInlineData<ProfileItemFragment$key>(ProfileItemFragment, profileRef)
```
anicioalexandre: why do we read from inline instead of using `useFragment`?

packages/components/modules/profiles/ProfilePopover/ProfilePopoverSections/index.tsx
anicioalexandre, line 1: that's a component we wont need after we refactor AccountPopover

packages/components/modules/profiles/ProfilePopover/ProfilesList/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        profileRef={profileRef}
+        currentProfile={currentProfile}
+        onProfileChange={handleProfileChange}
+        {...MenuItemProps}
+      />
+    )
+  })
+}
+
+const LoadingState: FC = () => (
```
anicioalexandre: let's avoid declaring multiple react components on the same file when we can

packages/components/modules/profiles/ProfilePopover/__tests__/ProfilePopover.cy.tsx
```diff
@@ -0,0 +1,294 @@
...
+    cy.contains(mockUserProfileData.data.me.profile.name).should('exist')
+    cy.contains(mockUserProfileData.data.me.profile.urlPath.path).should('exist')
+
+    const profileListData = mockProfilesListFactory(
+      profilesSize,
+      mockUserProfileData.data.me.profile,
+    )
+
+    if (openSwitchProfileList) {
+      cy.contains('button', /switch profile/i)
```
anicioalexandre: let's use `cy.findByRole('button', { name: /switch profile/i }) `selector instead

packages/components/modules/profiles/ProfilePopover/__tests__/ProfilePopover.cy.tsx
anicioalexandre, line 1: we could have used better (more accessible and less flaky) selectors here, also merged a could test cases into one

packages/components/modules/profiles/context/CurrentProfileProvider/index.tsx
anicioalexandre, line 1: agreed with @nossila , I'm afraid that's not what we want here...  useCurrentProfile should be very very simple (hold the current profile id and be able to change it) and nothing else. Any sort of API calls should've been done inside the components

packages/components/webpack.config.ts
```diff
@@ -40,6 +40,7 @@ const config: Configuration = {
               ],
               '@babel/preset-typescript',
             ],
+            plugins: [['@babel/plugin-transform-modules-commonjs', { loose: true }]],
```
anicioalexandre, line 43: why?

packages/components/modules/profiles/ProfilePopover/index.tsx
anicioalexandre, line 1: I don't think we need this component as well

anicioalexandre (changes requested):

packages/components/modules/profiles/graphql/queries/ProfilesList.ts
```diff
@@ -0,0 +1,11 @@
+import { graphql } from 'react-relay'
+
+export const ProfilesListQuery = graphql`
+  query ProfilesListQuery {
+    me {
+      profiles {
```
anicioalexandre, line 6: Although we not handling pagination on this story, it would be nice if we had paginated query already... @nossila don't we have any examples of that query on AIRTAB?

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
Hercilio1: There is no graphql requests in this AccountPopover component. It only gets the user from the JWL. I mocked the hook.

packages/components/modules/profiles/context/CurrentProfileProvider/index.tsx
Hercilio1: That's a component. Providers are components. The provider is optimized to make the request only once when the user signs in. There is no way of getting the default user profile without that request. That request must be there; otherwise, we would be triggering it more than once. I agree that this isn't the ideal size for it, but considering the complexity needed to ensure this is a good component, this was the best approach I could get to work. The feedback @nossila gave me was about the previous version of useCurrentProfile, which changed with this provider.

If you have a suggestion of a better approch considering that, lets do a quick sidebar.

packages/components/webpack.config.ts
Hercilio1: For mocking hooks this is needed. For testing, being able to mock hooks is really important.

packages/components/modules/profiles/ProfilePopover/index.tsx
Hercilio1: What do you mean? That was what the story requested, wasn't it?

packages/components/modules/profiles/graphql/queries/ProfilesList.ts
Hercilio1: I would need to cover that in the BE as well. I aligned with @nossila and IC, about that. We will need to create a separated story for handling the pagination of profiles.
That wasn't covered in the AC and since I'm already 200% over budget I think we can keep it like that for now and wait this new story.

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
Hercilio1: This renderItems accepts components (React.FC) or React.JSXElements.

The difference between them is that I cannot create React.FC components inside existing components (for reusing props, for example), whereas React.JSXElements can reuse the props.

That way, I can send simple elements such as `<h1>test</h1>` (which are React.JSXElements) or complex components (such as `<ProfileList />`).

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
Hercilio1: I wanted to reuse the component to avoid redundancy. I also didn't want to break compatibility with the previous version of it. I don't want to mix the domains of profiles and navigations. By using the controller, I didn't break compatibility, and one day, if I (or any developer with their respective projects) decide to remove the profiles component from the project and use the simple AccountPopover, they can do so easily.

I understand that we want to avoid overcomplicating components, but the only other way here is to have redundant code. I chose to make it more complex. However, if you think it's better, I can make it redundant. But I truly believe we shouldn't mix the domains, since I already avoided that.

packages/components/modules/navigations/Header/AccountMenu/CurrentUserPlaceholder/types.d.ts
Hercilio1: What would be a declaration? Isn't it a declaration?
```
export interface CurrentUserPlaceholderProps {
  user: User & JWTContent
}
```

packages/components/modules/profiles/ProfilePopover/ProfileMenuItem/index.tsx
Hercilio1: Inside of the useCurrentProfile I needed it because I couldn't use the hook (useFragment) anywhere (it's a lint rule, If I'm not mistaken). So I found this alternative. I can rollback and ignore the lint rule or create two fragments, one with the inline rule and the other with the useFragment hook. What do you think would be the best approch?

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
Hercilio1: When we have animations, yes. Any better approch that you know?

anicioalexandre: we can probably solve that by using cy.findBy... to locate elements

packages/components/webpack.config.ts
Hercilio1: As aligned, we decided to keep this plugin

packages/components/modules/profiles/context/CurrentProfileProvider/index.tsx
Hercilio1: As aligned, we decided to keep the provider as it's currently.

packages/components/modules/profiles/ProfilePopover/ProfileMenuItem/index.tsx
Hercilio1: As aligned, we decided to keep it like that for this first version.

packages/components/modules/profiles/graphql/queries/ProfilesList.ts
Hercilio1: As aligned, we will implement the pagination in a new story

deboracosilveira (approved):

packages/components/modules/profiles/ProfilePopover/SwitchProfileMenu/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { FC } from 'react'
+
+import { ChevronIcon } from '@baseapp-frontend/design-system'
+
+import { ButtonBase, MenuItem, Stack } from '@mui/material'
+
+import { PopoverContentBox } from '../../../navigations/Header/AccountMenu/AccountPopover/styled'
```
anicioalexandre: let's avoid reusing components from other modules like that. If it's something reusable between modules it should be placed at `packages/components/modules/__shared__`.
since that's a simple component (and it's also overridden with sx, it's better to create a styled just for the SwitchProfileMenu, at  `SwitchProfileMenu/styled.tsx`

packages/components/modules/profiles/ProfilePopover/CurrentProfile/index.tsx
```diff
@@ -0,0 +1,39 @@
+import { FC } from 'react'
+
+import { AvatarWithPlaceholder } from '@baseapp-frontend/design-system'
+
+import { Box, Typography } from '@mui/material'
+
+import { PopoverContentBox } from '../../../navigations/Header/AccountMenu/AccountPopover/styled'
```
anicioalexandre: let's avoid reusing components from other modules like that. If it's something reusable between modules it should be placed at `packages/components/modules/__shared__`.
since that's a simple component (and it's also overridden with sx, it's better to create a styled just for the SwitchProfileMenu, at  `CurrentProfile/styled.tsx`

when we create a styled component, we often want to put as many styles as we can inside it, to avoid having two styling sources (in this case the styled styles and the sx prop)

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/LogoutItem/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const LogoutItem: FC<LogoutItemProps> = ({
+  children,
+  handlePopoverOnClose,
+  logoutButtonLabel = 'logout',
+}) => {
+  const { logout } = useLogout()
+
+  return (
+    <PopoverContentBox display="flex" flexDirection="column" gap={0.5}>
```
anicioalexandre: `PopoverContentBox` is not very semantic since this is not a popover, I would create a styled under the `LogoutItem` folder or use  a `Box` instead

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/CurrentUser/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const CurrentUser: FC = () => {
+  const { user } = useJWTUser<User & JWTContent>()
+
+  if (!user) {
+    return null
+  }
+
+  return (
+    <PopoverContentBox sx={{ px: 1, py: 0.75 }}>
```
anicioalexandre: same here, `PopoverContentBox` is not very semantic since this is not a popover, I would create a styled under the `CurrentUser` folder or use a Box instead

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__storybook__/static/profile-1.png
anicioalexandre, line 1: let's add a comment TODO: BA-1744: Technical Debt - Switch Profiles

packages/components/.storybook/decorators/withProviders.tsx
```diff
@@ -15,18 +15,14 @@ const queryClient = new QueryClient()
...
-      resolveMostRecentOperation({ mockResolvers })
-    } catch (error) {
-      // eslint-disable-next-line
-      console.log(error)
-    }
-  }, [mockResolvers, resolveMostRecentOperation])
+  queueOperationResolver(mockResolvers)
 
   return (
     <QueryClientProvider client={queryClient}>
```
anicioalexandre, line 29: Need to add `<CurrentProfileProvider>` here as well, as it breaks a couple of the navigation storybooks

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/MenuItems/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { List, MenuItem, Stack } from '@mui/material'
+
+import { PopoverContentBox } from '../styled'
+import { MenuItemsProps } from './types'
+
+const MenuItems: FC<MenuItemsProps> = ({ handlePopoverOnClose, menuItems }) => {
+  if (!menuItems || !menuItems.length) return null
+  return (
+    <PopoverContentBox>
```
anicioalexandre: same here

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__storybook__/stories.tsx
```diff
@@ -1,37 +1,87 @@
...
+  decorators: [
+    withTokenSetup,
+    (Story, context) => {
+      localStorage.removeItem(CURRENT_PROFILE_STORAGE_KEY)
+
+      const { queueOperationResolver, queuePendingOperation } = context.parameters
+        .relayMockEnvironment as ReturnType<typeof createTestEnvironment>
+
+      queueOperationResolver(mockResolvers)
+      queuePendingOperation(ProfilesListQuery)
```
anicioalexandre: do we need this last one? doesn't seem to take any effect

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__storybook__/stories.tsx
```diff
@@ -1,37 +1,87 @@
...
+
+      return (
+        <CurrentProfileProvider>
```
anicioalexandre: I think we can remove that once we add it under `packages/components/.storybook/decorators/withProviders.tsx `

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/__tests__/AccountPopover.cy.tsx
anicioalexandre, line 1: let's add a comment TODO: BA-1744: Technical Debt - Switch Profiles

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,112 @@
...
-  }
+  const [openProfilesList, setOpenProfilesList] = useState(false)
 
-  const handleMenuItemClick = (onClick: () => void) => {
-    onClick()
+  const handlePopoverOnClose = () => {
     popover.onClose()
+    // If the profiles list is open, close it after the close animation.
+    if (openProfilesList) {
+      setTimeout(() => {
```
anicioalexandre: let's make sure we clear timeouts on the component's unmount

packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx
```diff
@@ -1,81 +1,123 @@
...
+  useEffect(
+    () => () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId)
+      }
+    },
+    [timeoutId],
+  )
+
+  const loadCurrentProfile: boolean = Boolean(CurrentProfile) && Boolean(profile)
```
anicioalexandre: `: boolean` is redundant here, typescript already infers the boolean type

anicioalexandre: I'll add that to the TODO's story

anicioalexandre (approved):

PR #124, opened 2024-10-28
BA-1721: Add Component Testing to the @baseapp-frontend Pipeline

luisguareschi:
Acceptance Criteria 
To ensure that both unit and component tests are executed as part of the pipeline, we will be updating the current setup to accommodate component testing. This involves renaming the existing test step to reflect that it only covers unit tests and adding a new step for component tests executed by Cypress.

Rename the existing "Test" step in GitHub Actions to "Unit Test" to clarify that it is only responsible for unit tests.

Update the existing unit test script on the `@baseapp-frontend` packages (utils and authentication) from `test` to `test:unit` to reflect its specificity.

Add a new npm script `"test:component"` in the `@baseapp-frontend/components` package.  

Ensure this script runs Cypress to perform headless component testing on all tests cases available inside the package

In the main GitHub Actions Main workflow, add both the `test:unit` and `test:component` scripts to the pipeline.

Ensure that component tests are run as a separate step following the unit tests. 

Notes 

Existent component test we have under @baseapp-frontend/components: https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/comments/Comments/__tests__/Comments.cy.tsx 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/36005

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
         run: pnpm lint
 
   test-unit:
-    name: Unit Test
+    name: Unit Test Applications and Packages
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
@@ -78,7 +78,7 @@ jobs:
         run: pnpm test:unit
 
   test-component:
-    name: Component Test
+    name: Component Test Applications and Packages
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,3 +76,26 @@ jobs:
 
       - name: Test Applications and Packages
         run: pnpm test:unit
+
+  test-component:
+    name: Component Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -75,4 +75,27 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Test Applications and Packages
```
anicioalexandre, line 77: let's change to `Unit Test Applications and Packages` and the component test to `Component Test Applications and Packages`

anicioalexandre (approved):

.github/workflows/main.yml
luisguareschi: @anicioalexandre Done!

priscilladeroode:
@luisguareschi and @anicioalexandre I think we need to generate a new version of the package, maybe a patch increase?

anicioalexandre:
> @luisguareschi and @anicioalexandre I think we need to generate a new version of the package, maybe a patch increase?

@priscilladeroode in that case, I dont think we need new versions, since most of the work was changing pipelines and altering test scripts, it won't really affect the packages code

priscilladeroode (approved):

PR #128, opened 2024-11-05
BA-1584 Add Storybook for Notifications

matheusysd:
- Add Storybook for notifications components:
  - **NotificationsPopover**
  - **NotificationsList (with notifications, loading state and empty state)**
  - **NotificationItem**

## Release Notes

- **New Features**
  - Introduced Storybook support for notifications components, enhancing documentation and testing.
  - Added new components: `NotificationItemWithQuery` and `NotificationsPopover` with respective stories.
  - Implemented mock data for various notification states, including empty and populated lists.
  - Added new entries for notifications in the Storybook configuration.

- **Bug Fixes**
  - Improved handling of mock resolvers and data in the `withProviders` function.

- **Documentation**
  - Updated changelog to reflect new version 0.0.15 and recent changes.


anicioalexandre (approved):

packages/components/.storybook/preview.ts
```diff
@@ -59,6 +59,10 @@ const preview: Preview = {
           // General
           'Logo',
           'Scrollbar',
+          // Notifications
```
deboracosilveira: This first list is related to the design-system package stories, since both storybooks are composed here. So you don't need to add stories here, since they are from components package and if you can, please change the name `storiesOrder` to `designSystemStoriesOrder`. Besides that, you can move //Notifications prior to // Social and `NotificationItem` before `NotificationsList`

packages/components/modules/notifications/NotificationsList/NotificationItem/__storybook__/mockResolvers.ts
```diff
@@ -0,0 +1,28 @@
...
+      id: 'UHJvZmlsZToxNTM=',
+    },
+    target: {
+      id: 'Q29tbWVudDoyMzc=',
+      __typename: 'Comment',
+    },
+    actionObject: {
+      id: 'Q29tbWVudDoyNDg=',
+      __typename: 'Comment',
+      body: 'This is the first comment reply.',
```
deboracosilveira, line 24: image
look like body is not appearing on the story

packages/components/modules/notifications/NotificationsPopover/__storybook__/stories.tsx
```diff
@@ -0,0 +1,104 @@
+import type { Meta, StoryObj } from '@storybook/react'
```
deboracosilveira, line 1: image
i think we need to mock the data for the drawer as well, now is just loading

deboracosilveira:
can we mock the "Mark all as read" action as well?

packages/components/modules/notifications/NotificationsPopover/__storybook__/stories.tsx
matheusysd: we decided to mock only the first call, and I think that's why we are not rendering the list. Now that we have `queueOperationResolver` I'll try using it

packages/components/modules/notifications/NotificationsList/NotificationItem/__storybook__/mockResolvers.ts
matheusysd: It seems we have some bug here, @anicioalexandre tried to help me, but we decided to left like this and work on a bug later

deboracosilveira (approved):

PR #132, opened 2024-11-12
BA-1776 [FE] restrict upload to <15 MB, always display input element

tsl-ps2:
- Restricts uploads to the dropzone to files <15 MB
- Always displays the dropzone input element. This fixes the following bug: Currently, after uploading an image the `Change Image` button has no effect: It tries to open the dropzone file input, but there is none.

Does anybody know where the original limitation to file sizes <15 MiB came from? Afaik, local storage is limited to 5 MiB. Surprisingly (?), I do not always get an error when trying to upload larger files. Anyway, I would also suggest to reduce the maximum file size.

## Summary by CodeRabbit

- **New Features**
	- Enhanced `Dropzone` component with notification capabilities for file size validation.
	- Users receive notifications if uploaded files exceed 15 MB.

- **Improvements**
	- Improved organization of the component's JSX structure for better readability.

### DIFF
--- a/packages/design-system/components/Dropzone/index.tsx
+++ b/packages/design-system/components/Dropzone/index.tsx
@@ -1,6 +1,6 @@
-import { FC, useState } from 'react'
+import React, { FC, useState } from 'react'
 
-import { getImageString } from '@baseapp-frontend/utils'
+import { getImageString, useNotification } from '@baseapp-frontend/utils'
 
 import { Box, Button, Card, Typography } from '@mui/material'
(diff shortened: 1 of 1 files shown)


tsl-ps2:
Does anybody know where the original limitation to file sizes <15 MiB came from? Afaik, local storage is limited to 5 MiB. Surprisingly (?), I do not always get an error when trying to upload larger files. Anyway, I would also suggest to reduce the maximum file size.

PR #135, opened 2024-11-13
BA-1777: multiple profiles member list

hpmoreira05:
As a user, on the BaseApp Profile,I would like to I want to have a page where I can see the members of a profile, In order to be able to manage the members.

 

Original Story:  
Acceptance Criteria 
Context
This story will be about creating the option to access the members page, creating the structure of the members page and also creating the members modal and the relationship with the profiles

Business Rules - Members List - 5

Given a user has navigated to the Main Profile Settings Page for any profile, then they should see the "Members" option in the side menu.

After clicking the Members option, the user should be able to see a list of the members for the active profile with the following properties

Show a count of the total members (active and pending invite)

Pending invite members should be displayed in a light gray font with a pending status.

The members list will have its scrolling behavior inside its component, this means the page will not scroll if you do it inside the member list component. 

Current behavior 

Expected behavior 

Code Snippet 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3660-64822&node-type=frame&t=MBvMfObBLgXLqFQv-0 

Template: https://bitbucket.org/silverlogic/baseapp-frontend-template/pull-requests/113
BA BE Package:  https://github.com/silverlogic/baseapp-backend/pull/182

## Release Notes

- **New Features**
  - Introduced `MemberItem` component for displaying member information.
  - Added `MembersList` component for paginated member display.
  - Implemented `MemberListItem` component for rendering individual member details in a list.
  - New GraphQL queries and fragments for user members and member items.

- **Enhancements**
  - Styled components for improved layout and visual representation of member items.
  - TypeScript interfaces added for better type safety and integration.

- **Constants**
  - Added constants for member loading configurations and status representations.

- **Utility**
  - New utility function to capitalize the first letter of strings.

- **Documentation**
  - Updated schema with new types and structures for enhanced functionality.

- **Version Update**
  - Updated package version from `0.0.22` to `0.0.23`.


packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const renderLoadingState = () => {
+    if (!isLoadingNext) return <Box sx={{ paddingTop: 3 }} />
+
+    return (
+      <LoadingState
+        sx={{ paddingTop: 3, paddingBottom: 1 }}
+        CircularProgressProps={{ size: 15 }}
+        {...LoadingStateProps}
+      />
+    )
```
Hercilio1, line 43: Instead of using paddingTop and paddingBottom, you can use, respectively, pt and pb. This alisases, inside of the sx props, make the code cleaner.

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const renderMemberItem = (member: any, index: number) => {
+    if (!member) return null
+    if (
+      member.status === MemberStatuses.active &&
+      members[index - 1]?.status !== MemberStatuses.active
+    ) {
+      return (
+        <>
+          <Divider />
+          <MemberItem member={data.profile} memberRole="owner" status={MemberStatuses.active} />
```
Hercilio1: You can add an if statement inside the return to avoid repeating the following component:
```
<MemberItem
            member={member?.user?.profile}
            memberRole={member?.role}
            status={member?.status}
          />
```

packages/components/modules/profiles/Members/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  <>
+    {Array.from({ length: NUMBER_OF_MEMBERS_ON_FIRST_LOAD }).map((_, index) => (
+      <MemberItemSkeleton
+        key={index} // eslint-disable-line react/no-array-index-key
+        variant="rectangular"
+        sx={{ mb: 0.5 }}
+      />
+    ))}
+  </>
+)
```
Hercilio1: The BA packages avoid having two different FCs in the same file. I suggest you to create a component file for the loading state.

packages/components/modules/profiles/Members/utils.ts
```diff
@@ -0,0 +1,2 @@
+export const capitalizeFirstLetter = (string: string) =>
```
Hercilio1, line 1: Why don't you use css instead?
`text-transform: capitalize;`

packages/components/schema.graphql
```diff
@@ -1,26 +1,42 @@
-"""Exposes a URL that specifies the behaviour of this scalar."""
```
Hercilio1: All comments were updated. I think that's an IDE configuration issue. I guess you didn't change this in propose, so I would revert these unneeded changes.

Hercilio1 (changes requested):

packages/components/modules/profiles/Members/MemberItem/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { AvatarProps, BoxProps } from '@mui/material'
+
+import type { ProfileItemFragment$key } from '../../../../__generated__/ProfileItemFragment.graphql'
+import type {
+  ProfileRoleStatus,
+  ProfileRoles,
+} from '../../../../__generated__/UserMembersListFragment.graphql'
+
+export interface IMemberPersonalInformation extends BoxProps {
```
anicioalexandre: let's avoid adding the I prefix

packages/components/modules/profiles/Members/MembersList/index.tsx
anicioalexandre: I think that's arguable, it looks more readable without the alisases

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+      />
+    )
+  }
+
+  const renderMemberItem = (member: any, index: number) => {
+    if (!member) return null
+    if (
+      (member.status === MemberStatuses.active &&
+        members[index - 1]?.status !== MemberStatuses.active) ||
+      (member.status === MemberStatuses.active && !members[index - 1]?.status)
```
anicioalexandre: could we move some of these conditions to a constant? just so we can read better what the rule is, something like:
```
    const isActiveMember = member.status === MemberStatuses.active
    const isPreviousMemberInactive = members[index - 1]?.status !== MemberStatuses.active
    const isPreviousMemberUndefined = !members[index - 1]?.status

    const isFirstActiveMember = (isActiveMember && isPreviousMemberInactive) || (isActiveMember && isPreviousMemberUndefined)
```

packages/components/modules/profiles/Members/index.tsx
anicioalexandre: @hpmoreira05 let's move that to `modules/profiles/Members/DefaultInitialLoadingState`

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          1 member
+        </Typography>
+        <MemberItem member={data} memberRole="owner" status={MemberStatuses.active} />
+      </>
+    )
+  }
+
+  return (
+    <>
+      {data.members?.totalCount && (
```
anicioalexandre: I think we could simplify that a little bit
      <Typography variant="subtitle2" mb={4}>
```suggestion
      <Typography variant="subtitle2" mb={4}>
        {data.members?.totalCount ?? 1} members
      </Typography>
```

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { Virtuoso } from 'react-virtuoso'
+
+import { UserMembersListFragment } from '../../graphql/queries/UserMembersList'
+import DefaultMemberItem from '../MemberItem'
+import { MemberStatuses, NUMBER_OF_MEMBERS_TO_LOAD_NEXT } from '../constants'
+import { MemberListProps } from '../types'
+
+const MembersList: FC<MemberListProps> = ({
+  userRef,
+  MemberItem = DefaultMemberItem,
```
anicioalexandre, line 18: we're missing MemberItemProps here

packages/components/modules/profiles/Members/index.tsx
anicioalexandre: I would make a simpler loading state, maybe just using the <LoadingState /> component, and wait for the designers to have a skeleton, maybe on the next versions

packages/components/modules/profiles/Members/MemberItem/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+          {...avatarProps}
+        />
+        <Box>
+          <Typography variant="subtitle2">{memberProfile.name}</Typography>
+          <Typography variant="caption">{memberProfile?.urlPath?.path}</Typography>
+        </Box>
+      </MemberPersonalInformation>
+
+      <Box>
+        <Button variant="soft" color="inherit">
```
anicioalexandre: let's remove the pointer-events here since it should not be interactive `sx={{pointerEvents: "none"}}`

anicioalexandre (commented):
that's a very good PR! congrats 🎉

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    return (
+      <LoadingState
+        sx={{ paddingTop: 3, paddingBottom: 1 }}
+        CircularProgressProps={{ size: 15 }}
+        {...LoadingStateProps}
+      />
+    )
+  }
+
+  const renderMemberItem = (member: any, index: number) => {
```
anicioalexandre: we can avoid the `any` by typing it like this:
```
// on Members/types.ts
type Members = NonNullable<UserMembersListFragment$data['members']>
type MembersEdges = Members['edges']
export type MemberNode = NonNullable<MembersEdges[number]>['node']

// on MembersList/index.tsx
const renderMemberItem = (member: MemberNode, index: number) => {
```

packages/components/modules/profiles/Members/MemberItem/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const MemberItem: FC<MemberItemProps> = ({
+  member,
+  memberRole,
+  status,
+  avatarProps = {},
+  avatarWidth = 40,
+  avatarHeight = 40,
+}) => {
+  const memberProfile = useFragment<ProfileItemFragment$key>(ProfileItemFragment, member)
```
anicioalexandre, line 23: I think ideally this fragment should be related to members (aka ProfileUserRole on the backend) instead of profile fragment, we would use a MembersItemFragment. So the list query would be something like this:
```
export const UserMembersListFragment = graphql`
  fragment UserMembersListFragment on Profile
  @refetchable(queryName: "userMembersListPaginationRefetchable")
  @argumentDefinitions(
    count: { type: "Int", defaultValue: 10 }
    cursor: { type: "String" }
    orderByStatus: { type: "String", defaultValue: "custom" }
  ) {
    members(first: $count, after: $cursor, orderByStatus: $orderByStatus)
      @connection(key: "UserMembersFragment_members", filters: ["orderByStatus"]) {
      totalCount
      edges {
        node {
          ...MemberItemFragment
        }
      }
      pageInfo {
        endCursor
        hasNextPage
      }
    }
  }
`
```

them the members fragment something like:
```
export const MemberItemFragment = graphql`
fragment MemberItemFragment on Member // need to create this interface
@argumentDefinitions(avatarSize: { type: "Int", defaultValue: 100 }) {
  id
  user {
    profile {
      ...ProfileItemFragment
    }
  }
  role
  status
}
`
```
@nossila what do you think? I think we mentioned that on our last sidebar

anicioalexandre: then if we do that, we avoid needing to pass over the `member?.role` and `member?.status` since we would define those fields on the MemberItem fragment

packages/components/modules/profiles/Members/MemberListItem/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { MemberStatuses } from '../constants'
+import { MemberListItemProps } from './types'
+
+const MemberListItem: FC<MemberListItemProps> = ({
+  member,
+  data,
+  prevMember,
+  nextMember,
+  MemberItemComponent,
+  memberItemComponentProps,
```
anicioalexandre, line 16: let's use capital letter for these kind of props

packages/components/modules/profiles/Members/MembersList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <>
+      <Typography variant="subtitle2" mb={4}>
+        {(data.members?.totalCount ?? 0) + 1} members
+      </Typography>
+      <Virtuoso
+        style={{ height: membersContainerHeight }}
+        data={members}
+        overscan={NUMBER_OF_MEMBERS_TO_LOAD_NEXT}
```
anicioalexandre: lets not overscan for now, on large lists can cause Maximum call stack size exceeded error

packages/components/modules/profiles/Members/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const Members: FC<UserMembersProps> = ({
+  MemberItem,
+  LoadingState,
+  LoadingStateProps,
+  membersContainerHeight,
+}) => {
+  const { profile: currentProfile } = useCurrentProfile()
+
+  const data = useLazyLoadQuery<IUserMembersListPaginationQuery>(UserMembersListPaginationQuery, {
```
anicioalexandre, line 24: lets avoid the I prefix for type interfaces

anicioalexandre (approved):

PR #145, opened 2024-12-01
BA-1832: organizations

priscilladeroode:
As a user, on the BaseApp Profile,I would like to Create an organization profile, In order to To create social activites in the name of that organization profile.

 
Acceptance Criteria 
Context
This story will handle creating a new organization profiles and also setting the logic for the relationship between members and organizations. 

Business Rules - Create Organization

Given I select the Switch Profile option in my User Settings dropdown, then display an [Add new organization] option at the bottom of the list.

After clicking that new option, a modal should appear with the following inputs:

Name

URL path  - this will be the details to navigate to the profile in the URL. 

(Ex: airtab.com/BigAlex)

Given the user has submitted the form, display a toast indicating the organizations creation was succesful.

Given the creation was succesful, then the user that created the organization should be set as a member with the Owner Role

The user should now see this organization in his Switch Profile list.

Validation Rules

The username should be unique, meaning that we should display an error message if the username has already been taken.

Both input fields should have content before enabling the "Create organization button"

Copy Text

Username already in use, please choose another one" 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=218-293195&node-type=canvas&t=o7CIQdSvtnKDiy3z-0 

Notes 
Implementation Rules

Create Organization modal

Needs to inherent from the baseapp ProfieableModel

Create GraphQL mutation 

Approvd 
https://app.approvd.io/projects/BA/stories/34196 

## Release Notes

- **New Features**
  - Introduced a modal for creating new organization profiles.
  - Added functionality to manage profile visibility and permissions dynamically.
  - Enhanced the `ProfilesList` component with pagination and loading states.
  - Implemented a new query for fetching user permissions related to organization creation.
  - Added a new styled component for form layouts.
  - Introduced a new constant for URL path validation.

- **Bug Fixes**
  - Improved error handling for profile data fetching.
  - Enhanced safety in accessing profile data to prevent runtime errors.

- **Documentation**
  - Updated GraphQL schema with new types and improved comments for clarity.

- **Chores**
  - Updated dependencies for better version control and stability.
  - Added new scripts for managing GraphQL schema.


packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+const schema = z.object({
+  name: z
+    .string()
+    .trim()
+    .min(3, { message: 'Must have at least three character' })
+    .max(255, { message: 'Must have at most 255 characters' }),
+  urlPath: z
+    .string()
+    .regex(slugRegex, 'Must contain only lowercase letters, numbers, and hyphens')
```
anicioalexandre: Do the same rules apply to the backend as well?

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+  DialogTitle,
+  Link,
+  Typography,
+} from '@mui/material'
+import { useForm } from 'react-hook-form'
+import { ConnectionHandler } from 'react-relay'
+import slugify from 'slugify'
+import z from 'zod'
+
+import { Form } from '../../../__shared__/SocialInput/styled'
```
anicioalexandre: shouldn't this `Form` come from `AddProfileModal/styled`?

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+  addNewProfileLabel?: string
+  termsAndConditionsUrl?: string
+  addNewProfileDescription?: string
+  onClose?: () => void
+  open: boolean
+  setOpen: (open: boolean) => void
+  submitLabel?: string
+}
+
+interface OrganizationCreateForm {
```
anicioalexandre: let's move the type interfaces to `AddProfileModal/types`

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+import { ConnectionHandler } from 'react-relay'
+import slugify from 'slugify'
+import z from 'zod'
+
+import { Form } from '../../../__shared__/SocialInput/styled'
+import { useOrganizationCreateMutation } from '../../graphql/mutations/OrganizationCreate'
+
+const slugRegex = /^[a-z0-9]+(?:-[a-z0-9]+)*$/
+
+const schema = z.object({
```
anicioalexandre: let's move these schema constants to a `AddProfileModal/constants` file

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+    commitMutation({
+      variables: {
+        input: {
+          name: data.name,
+          urlPath: data.urlPath,
+          clientMutationId,
+        },
+      },
+      updater: (store) => {
```
anicioalexandre: we probably don't need an `updater` function for this case.  We could try to use `@appendEdge` or `@prependEdge` in order to add the new edge to the list, so on the mutation query it would be:

```
export const OrganizationCreateMutationQuery = graphql`
  mutation OrganizationCreateMutation($input: OrganizationCreateInput!, $connections: [ID!]!) {
    organizationCreate(input: $input) {
      organization @prependEdge(connections: $connections) {
        node {
          id
        }
      }
      profile {
        node {
          ...ProfileItemFragment
        }
      }
      errors {
        field
        messages
      }
    }
  }
`
```

but for that, we need to make sure we have a `connection` on the query (which already has `@connection(key: "ProfilesListFragment_profiles") `). So now we just need to make sure we pass the connection id on the commit function, something like:
```
   const connectionID = ConnectionHandler.getConnectionID(roomId, 'ProfilesListFragment_profiles')
    
    commitMutation({
      variables: {
        input: {
          name: data.name,
          urlPath: data.urlPath,
          clientMutationId,
        },
        connections: [connectionID],
      }
```

[adding edges docs](https://relay.dev/docs/guided-tour/list-data/updating-connections/#using-declarative-directives)
[connections docs](https://relay.dev/docs/guided-tour/list-data/connections/)

nossila: @anicioalexandre exactly! just point that the `connectionId` should be `const connectionID = ConnectionHandler.getConnectionID(user.id, 'ProfilesListFragment_profiles')`. `user.id` instead of `roomId`

packages/components/modules/profiles/context/CurrentProfileProvider/__tests__/CurrentProfileProvider.test.tsx
```diff
@@ -102,6 +102,7 @@ describe('CurrentProfileProvider', () => {
 
     await waitFor(() => {
       const storedData = JSON.parse(localStorage.getItem(CURRENT_PROFILE_STORAGE_KEY) || '{}')
+      console.log('storedData------', storedData)
```
deboracosilveira: remove

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
priscilladeroode: Updated @nossila and @anicioalexandre

packages/components/modules/profiles/ProfilePopover/AddProfileMenuItem/index.tsx
```diff
@@ -1,20 +1,48 @@
...
+          <AddIcon color="action" />
+        </MenuItem>
+      </Stack>
+      <CreateProfileModal open={open} setOpen={setOpen} userId={me?.id} />
+    </>
+  )
+}
+
+const AddProfileMenuItemSuspended: FC<AddProfileMenuItemProps> = (props) => (
+  <Suspense>
```
anicioalexandre, line 43: could we add a` TODO: add a fallback loading state` here?

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+      aria-labelledby="organization-modal-title"
+      aria-describedby="organization-modal-description"
+      PaperProps={{
+        sx: {
+          borderRadius: { xs: 0, sm: 2 },
+          minWidth: { xs: '100%', sm: 0 },
+          margin: { xs: 0, sm: 2 },
+          maxWidth: { sm: 366 },
+          height: { xs: '100%', sm: 'auto' },
+          minHeight: { xs: '100%', sm: 'auto' },
```
anicioalexandre, line 103: is that specified in the designs? I'm asking because we do have a dialog component under `packages/design-system/components/dialogs/Dialog` so if that's a style that applies to all our dialogs, we should consider doing these changes on that component

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+        >
+          <Typography color="text.secondary">{addNewProfileDescription}</Typography>
+          <Box display="flex" flexDirection="column" gap={2}>
+            <TextField
+              label="Name"
+              name="name"
+              size="medium"
+              control={form.control}
+              onChange={() => {
+                form.trigger('name')
```
anicioalexandre: just curious why we have this trigger, if we set `useForm({mode: "onChange"})` wouldn't that be enough?

packages/components/modules/profiles/ProfilePopover/AddProfileModal/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+          </Button>
+          <LoadingButton
+            variant="contained"
+            color="inherit"
+            sx={{ width: 'auto' }}
+            type="submit"
+            loading={isMutationInFlight}
+            disabled={
+              isMutationInFlight ||
+              Object.keys(form.formState.errors).length > 0 ||
```
anicioalexandre, line 157: would it be a case where if `formState.isValid` is true there would still be `formState.errors`? asking because we normally only use isValid

anicioalexandre (approved):

PR #154, opened 2024-12-10
BA-1917: Archive Rooms

pt-tsl:
**Acceptance Criteria**
Business rules - Archiving and Archive Chat List
As a user I want to be able to archive selected chats and access them in a dedicated "Archived Chat List" so that I can keep my active chats organized while preserving old conversations for future reference.

When I click on the "Archive" button on a chat card overlay, the selected chat room is removed from my active chat list and moved to the "Archived Chat List."

When I navigate to the "Archived Chat List" section, I can view all archived chat rooms.

The Archived Chat rooms should be sorted by the timestamp in which they were archived. 

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for archiving and unarchiving chat rooms.
	- Added new `UnarchiveIcon` to the design system.
- **Enhancements**
	- Improved state management for unread and archived chat rooms in the ChatRoomsList component.
	- Enhanced ChatRoomItem component to handle archived and unread states.
	- Updated RoomsListFragment to support filtering of archived chat rooms.
	- Upgraded dependencies, including the design system to version 0.0.24.
- **Bug Fixes**
	- Updated mutation handling for better management of chat room states.
- **Documentation**
	- Updated changelog to reflect version 0.0.32 and the new features included.


packages/components/modules/messages/ChatRoomsList/ChatRoomItem/index.tsx
```diff
@@ -53,16 +57,43 @@ const ChatRoomItem: FC<ChatRoomItemProps> = ({
...
           label: 'Archive Chat',
-          onClick: () => {},
+          onClick: () => {
+            if (currentProfile?.id) {
+              commit({
+                variables: {
+                  input: {
+                    roomId: room.id,
+                    profileId: currentProfile.id,
+                    archive: !isInArchivedTab,
```
tsl-ps2, line 83: Why do you need to pass this argument? Is this also unarchiving the chat (with the same archive icon)?

packages/components/modules/messages/ChatRoomsList/ChatRoomItem/index.tsx
```diff
@@ -53,16 +57,43 @@ const ChatRoomItem: FC<ChatRoomItemProps> = ({
...
+                  },
+                },
+                updater: (store: RecordSourceSelectorProxy<unknown>) => {
```
tsl-ps2: This does not depend on the server response. So it always removes the chat, even if there was an error on the backend? Or am I missing something?

packages/components/modules/messages/ChatRoomsList/ChatRoomItem/index.tsx
tsl-ps2: I see, it is there because of the next PR.

packages/components/modules/messages/ChatRoomsList/ChatRoomItem/index.tsx
pt-tsl: If I'm not mistaken, relay doesn't update if there is a server error. But I don't think it would hurt to include error handling in that function, to be safe

anicioalexandre (approved):

anicioalexandre (approved):

PR #155, opened 2024-12-11
BA-1940: unarchive chat

pt-tsl:
**Acceptance Criteria**
**Business Rules - Archive Reversibility**
As a user , I want to be able to unarchive a previously archived chat so that I can bring the conversation back to my active chat list if it becomes relevant again.

- When I hover my mouse over a chat list card in the Archive Chat list, an overlay with an "UnArchive" icon or button appears on the card.
- When I hover over a chat card, the overlay appears within one second to ensure it is quick and responsive.
- Also include the  "Hide (eye)" and "Delete" icon, but will not be functional for now
- After clicking on the "Unarchive button" the chat room should appear in the Active Chat List, following the sorting that the Active Chat list already has (based on the timestamp of the last message).

***DESIGNS ARE PENDING**

Change the "Archive" button so that users identify it will unarchive the chat room.


packages/components/modules/messages/ChatRoom/index.tsx
```diff
@@ -32,6 +35,34 @@ const ChatRoom: FC<ChatRoomProps> = ({
...
     },
   )
+  const { currentProfile } = useCurrentProfile()
+  const [commitMutation] = useReadMessageMutation()
+
+  const prevProfileIdRef = useRef<string | undefined>(currentProfile?.id)
+  const prevRoomIdRef = useRef<string | undefined>(chatRoom?.id)
+
+  useEffect(() => {
+    if (hasRunRef.current)
```
tsl-ps2: Would this break in non-strict mode? But if I understood you right, you will remove this in the next PR anyway, so it does not matter.

packages/components/modules/messages/ChatRoomsList/ChatRoomItem/index.tsx
```diff
@@ -53,16 +62,43 @@ const ChatRoomItem: FC<ChatRoomItemProps> = ({
...
         {
-          disabled: true,
-          icon: <ArchiveIcon />,
-          label: 'Archive Chat',
-          onClick: () => {},
+          disabled: isMutationInFlight,
+          icon: !isInArchivedTab ? <ArchiveIcon /> : <UnarchiveIcon />,
+          label: !isInArchivedTab ? 'Archive Chat' : 'Unarchive Chat',
+          onClick: () => {
+            if (currentProfile?.id) {
```
tsl-ps2, line 77: I personally like to extract such pieces of code into functions to keep the render method shorter, but feel free to ignore this!

anicioalexandre (approved):

PR #171, opened 2025-01-03
BA-2035: admin permissions

tsl-ps2:
**Description**

- As a user, on the BaseApp Messages,I would like to Have admin permissions when I create a group, In order to Add/remove members and change group details.

**Original Story:**  
Acceptance Criteria 
Admin Permissions 

- Given I am the user who created the group, I should have admin permissions.
- Given I am the group admin, when I view the group details, then I should have the ability to:
  - Add or remove members.
  - Change group settings (name, profile picture,)
- Given I am in the group details page, I should see a 3 dots menu next to the members in the list, when I click on that 3 dots menu then I should see the options:
  - Make member an admin
  - Remove member

**Guardrails**

- We will not implement removing a member or making a member an admin in this story. Only rendering the 3 dots menu.


packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx
```diff
@@ -2,19 +2,29 @@
...
 import { MainContainer } from './styled'
 import { ProfileCardProps } from './types'
 
-const ProfileCard: FC<ProfileCardProps> = ({ profile: profileRef, role }) => {
+const ProfileCard: FC<ProfileCardProps> = ({ hasAdminPermissions, profile: profileRef, role }) => {
   const { id, image, name, urlPath } = useFragment(ProfileItemFragment, profileRef)
-  const showPath = !!urlPath?.path
-  const showAdmin = role === 'ADMIN'
+  const showUrlPath = !!urlPath?.path
+  const showAdminLabel = role === 'ADMIN'
```
pt-tsl: Maybe move the `ADMIN` const to a dict with all possible roles

packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx
```diff
@@ -48,10 +58,25 @@ const ProfileCard: FC<ProfileCardProps> = ({ profile: profileRef, role }) => {
             />
           )}
           <Typography variant="caption" color="primary.light">
-            {showAdmin && 'Admin'}
+            {showAdminLabel && 'Admin'}
```
pt-tsl: Same as above comment

pt-tsl (changes requested):
Check comments in packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx

packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx
tsl-ps2: I created an enum for all possible roles. Since this is only the label that we display (the only other role is regular member which is not displayed), I don't think it makes much sense to create a dictonary/enum for the labels? And having an extra constant `ADMIN_LABEL = 'Admin'` looks a bit superfluous to me, so I would leave it as is?

packages/components/modules/messages/GroupDetails/ProfileCard/constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum ChatRoomParticipantRoles {
```
anicioalexandre: would be cool to not use anums and prefers a constant + type instead
```suggestion
import { ValueOf } from '@baseapp-frontend/utils'

export const CHAT_ROOM_PARTICIPANT_ROLES = {
  admin: 'ADMIN',
  member: 'MEMBER',
} as const

export type ChatRoomParticipantRoles = ValueOf<typeof CHAT_ROOM_PARTICIPANT_ROLES>
```

anicioalexandre (changes requested):

packages/components/modules/messages/GroupDetails/ProfileCard/constants.ts
anicioalexandre: enums can have some tradeoffs and little [bugs](https://github.com/microsoft/TypeScript/issues?q=is%3Aissue+is%3Aopen+enum+label%3Abug), constants + types offers more stability and doesnt need to be transpilled to code

tsl-ps2: Good suggestion, thanks!

packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx
pt-tsl: I honestly don't think it's superfluous to have a dictionary for labels, we usually have them. But I'm fine with not having one now.

pt-tsl (approved):

packages/components/modules/messages/GroupDetails/ProfileCard/index.tsx
tsl-ps2: If we usually have them, I will add it.

tsl-ps2:
Will merge by hand into (a rebased version of) BA-2024-group-details-page

tsl-ps2:
Will merge by hand into (a rebased version of) BA-2024-group-details-page

PR #173, opened 2025-01-08
BA-2054 [components] Removing a member

tsl-ps2:
**Short demo**

[https://www.loom.com/share/6115a6fa1ab1457da64797f84edeab31](https://www.loom.com/share/6115a6fa1ab1457da64797f84edeab31)

**Notes**

This includes the changes to the graphql files that we discussed during breakouts.

**Description**

As a user, on the BaseApp Messages,I would like to Remove an existing member of the group, In order to avoid that person to have access to the group messages.

**Acceptance Criteria 
Removing a Member**

- Given I am an admin of the group, when I select a member from the group details view, then I should have the option to remove that member from the group.

  - We should have already implemented the Remove Member option button in the Admin Permission story, but not the functionality. 

- Given I click on remove member, I should see a confirmation dialog, to make sure I want to do this action.

- Given I confirm the removal of the member, then the removed member should no longer see the group in their chat list or have access to its messages.

  - We are doing a hard delete of the group chat for that user. 

## Release Notes

- **New Features**
  - Enhanced chat room management with improved participant tracking.
  - Added support for more granular role management in chat rooms (Admin/Member roles).
  - New GraphQL fragments for Last Message, Title, and Unread Messages Count to streamline data retrieval.
  - Introduced new properties for chat room components to enhance data handling.
  - Added `participantsCount` to chat room queries and fragments for better participant management.

- **Improvements**
  - Refined chat room header and title information retrieval.
  - Enhanced participant count and removal tracking.
  - Updated GraphQL schema to support more detailed chat room interactions.
  - Improved subscription management for real-time updates in chat rooms.
  - Enhanced flexibility in handling connections within chat room subscriptions.

- **Bug Fixes**
  - Improved handling of chat room connections.
  - More robust error handling for room and participant management.

- **Performance**
  - Optimized GraphQL data fetching with more specific fragments.
  - Improved connection handling in Relay environment.


packages/components/modules/messages/graphql/subscriptions/useRoomListSubscription.tsx
```diff
@@ -1,41 +1,110 @@
...
+      },
+      onNext: (data: useRoomListSubscription$data | null | undefined) => {
+        if (
+          wasRemovedFromChatRoom(data) &&
+          selectedRoom &&
+          data?.chatRoomOnRoomUpdate?.room?.node?.id === selectedRoom
+        ) {
+          resetChatRoom()
+          onRemoval?.()
+          sendToast(`You were removed from ${data?.chatRoomOnRoomUpdate?.room?.node?.title}`, {
```
pt-tsl, line 97: As it is, this toast won't be displayed to the removed user. Also, I wasn't able to find this toast on the design or on the AC, are we sure this is something we should be implementing?

pt-tsl (changes requested):
Left a simple comment, will continue to check

packages/components/modules/messages/graphql/subscriptions/useRoomListSubscription.tsx
tsl-ps2: It was displayed but only if you had the chat room open in the right panel. I switched to always displaying it.

pt-tsl (approved):

Ronan-Fernandes (approved):

packages/components/modules/messages/graphql/fragments/Title.ts
```diff
@@ -0,0 +1,15 @@
+import { graphql } from 'react-relay'
+
+// Ideally this would use a type refinement of the form
+// ...on ChatGroup { ...GroupTitleFragment }
+// ...on OneOnOneChat { ...RoomTitleFragment }
```
anicioalexandre, line 5: I think we could do it on `baseapp-chats/baseapp_chats/graphql/object_types.py`, @nossila do you think it makes sense?

packages/components/modules/messages/GroupDetails/index.tsx
```diff
@@ -1,45 +1,55 @@
...
 import { GroupDetailsHeader } from './GroupDetailsHeader'
 import DefaultProfileCard from './ProfileCard'
 import { CHAT_ROOM_PARTICIPANT_ROLES } from './ProfileCard/constants'
+import { GroupMembersEdge } from './ProfileCard/types'
 import { GroupHeaderContainer, GroupTitleContainer } from './styled'
-import { GroupDetailsProps, GroupMembersEdge } from './types'
+import { GroupDetailsProps } from './types'
 
-const GroupDetails: FC<GroupDetailsProps> = ({
+const GroupDetails: FC<GroupDetailsProps & { profileId: string }> = ({
```
anicioalexandre: let's just update `GroupDetailsProps` directly, if we really need to pass the current profile id as props

anicioalexandre (approved):

packages/components/modules/messages/GroupDetails/index.tsx
tsl-ps2: This is a bit awkward, the issue is that you cannot call hooks conditionally. The `useSubscription` hook needs a non-null `profileId`, but since it is a hook, it cannot perform that check (profileId is non-null) itself and neither can the component calling the hook. This is why `GroupDetails` is wrapped below to provide a non-null id, the wrapping component is exported and has `GroupDetailsProps` as props.

tsl-ps2: @anicioalexandre: I didn't see any other easier way to do this. We could also use `requestSubscription` instead, but then we need to make sure that cancelling the subscription works properly (which is why I didn't want to do this), use a package which provides a conditional `useSubscription` (needs an extra package just for handling a non-null id), or use `useSubscription` with a custom fetch function that does nothing if the `profileId` is null (a bit of a hack). I didn't like any of these options.

PR #176, opened 2025-01-09
BA-2060: leave group chat

pt-tsl:
**Acceptance Criteria** 
_Leave group - As a Member_

- Given I am a member of a Group shat, I should see a 3 dots menu in the Chat room title area.
- Given I am a member of a group chat, when click the 3 dots menu in the Group Title, then I should see a "Leave Group" option prominently displayed.
- Given I click on the "Leave Group" button, when the action is triggered, then I should see a confirmation prompt asking if I am sure I want to leave the group.
- Given I confirm that I want to leave the group, then I should no longer see the group in my chat list and stop receiving updates of messages on that group.

**We are going to do a hard delete for that Group Chat for that user.**

## Summary by CodeRabbit

- **New Features**
	- Added chat room options menu with actions for archiving, viewing details, and leaving group
	- Introduced new dialogs for group management and participant removal
	- Enhanced profile card with more flexible menu options

- **Improvements**
	- Streamlined user interaction in group chat settings
	- Updated type definitions for better type safety
	- Simplified component logic for menu rendering

- **Bug Fixes**
	- Improved handling of group chat participant management


packages/components/modules/messages/ChatRoom/ChatRoomHeader/index.tsx
```diff
@@ -1,72 +1,118 @@
...
-    </ChatHeaderContainer>
+            {roomHeader.isGroup && (
+              <Typography component="span" variant="caption" sx={{ float: 'left', clear: 'left' }}>
+                {members}
+              </Typography>
+            )}
+          </Box>
+          <Box>
+            <IconButton
+              onClick={(e) => onChatRoomOptionsClicked(e)}
```
tsl-ps2: `onClick={onChatRoomOptionsClicked}` is a bit shorter

packages/components/modules/messages/ChatRoom/ChatRoomHeader/index.tsx
```diff
@@ -1,72 +1,118 @@
...
+              open={popover.open}
+              onClose={popover.onClose}
+              slotProps={{
+                root: { onClick: (e: React.MouseEvent<HTMLDivElement>) => e.stopPropagation() },
+              }}
+            >
+              <ChatRoomOptions
+                onArchiveClicked={() => {}}
+                onDetailsClicked={() =>
+                  roomHeader.isGroup ? onDisplayGroupDetailsClicked : undefined
```
tsl-ps2: Does this really work without invoking `onDisplayGroupDetailsClicked`? Or should it rather be `onDetailsClicked={roomHeader.isGroup ? onDisplayGroupDetailsClicked : undefined}` (or `onDetailsClicked={() => roomHeader.isGroup ? onDisplayGroupDetailsClicked() : undefined}`)

packages/components/modules/messages/GroupDetails/ProfileCard/MemeberOptionsMenu/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { FC } from 'react'
```
tsl-ps2: Typo in filename `MemeberOptionsMenu`

packages/components/modules/messages/__shared__/LeaveGroupDialog/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const onRemoveConfirmed = () => {
+    if (!roomId || !profileId) return
+    commit({
+      variables: {
+        input: {
+          roomId,
+          profileId,
+          removeParticipants: [removingParticipantId ?? profileId],
+        },
+        connections: [ConnectionHandler.getConnectionID(roomId, 'ChatRoom_participants')],
```
tsl-ps2, line 36: Does this work if the connection does not exist (when you leave a group from the chat room header menu)? Then connections is probably `[undefined]` but the argument definition in relay is something like `[ID!]!` (i.e. array of non-zero ids).
If you know this works, just ignore the comment.

packages/components/modules/messages/__shared__/LeaveGroupDialog/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { LoadingButton } from '@mui/lab'
+import { ConnectionHandler } from 'react-relay'
+
+import { useUpdateChatRoomMutation } from '../../graphql/mutations/UpdateChatRoom'
+import { LeaveGroupDialogProps } from './types'
+
+const LeaveGroupDialog: FC<LeaveGroupDialogProps> = ({
+  title = 'Leave group chat?',
+  content = 'You will stop receiving messages from this conversation and people will see that you left.',
```
tsl-ps2, line 16: I see that this is the message from the designs, but probably the designs were written assuming that we do a soft delete? Given that we implemented a hard delete, I think a wording like 'You will no longer see nor receive messages from this conversion and people will see that you left.' would be more appropriate (at least the user should be warned that he can no longer access previous messages). So we should maybe double check whether this message is the intended one?

tsl-ps2:
Is there a backend PR for this? Otherwise, I am afraid that only the admin can leave a group that way. (When I wrote the `UpdateGroupChatMutation` on the backend, I only thought about removing a member, for which you need admin permissions, and did not take leaving groups into account.)

packages/components/modules/messages/__shared__/LeaveGroupDialog/index.tsx
pt-tsl: Apparently this got approved, so I'm keeping as is

packages/components/modules/messages/ChatRoom/ChatRoomHeader/index.tsx
pt-tsl: It works, but I can see how that can lead to confusion. I'll change it.

packages/components/modules/messages/__shared__/LeaveGroupDialog/index.tsx
pt-tsl: Not sure if I understood what you meant by "the connection does not exist (when you leave a group from the chat room header menu)". The connection exists in the MembersList fragment, so it doesn't really matter from where you invoke the connection: relay will just update that store and, if the component that uses that fragment is mounted, that update will trigger a rerender.

pt-tsl:
> Is there a backend PR for this? Otherwise, I am afraid that only the admin can leave a group that way. (When I wrote the `UpdateGroupChatMutation` on the backend, I only thought about removing a member, for which you need admin permissions, and did not take leaving groups into account.)

Tbh I overlooked that. Thanks for pointing it out, I'll check and see if any changes are required

Ronan-Fernandes (approved):

anicioalexandre (approved):

PR #183, opened 2025-01-14
BA-2084: Multiple Profiles - Search Bar

luisguareschi:
As a user, on the BaseApp Profile,I would like to use a searc bar, In order to find invited and current members on the member list.

 

Original Story:  
Acceptance Criteria 
Context
In this story we want to implement the search bar in the Members List page. 

Business Rules - Search Bar - 3

Users should be able to use a search bar to locate a member:

 The search should be case-insensitive, so typing “john” should return results for “John” or “john.”
**The search should support partial matches, meaning typing part of a contact's name (e.g., “Jo” for “John Doe”) should filter the list to show all matching results.

The search bar should include a clear button (e.g., an “X” icon) that, when clicked, clears the search input and restores the full members list

Searchbar component should already be done, so re use it. 

Current behavior 

Expected behavior 

Code Snippet 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3660-64822&node-type=frame&t=IBg7GNDkZEpy2W77-0 

Approvd 
https://app.approvd.io/projects/BA/stories/35054 

## Release Notes

- **New Features**
  - Added search functionality for the members list.
  - Introduced a search bar for filtering members by first name, last name, and email.
  - Enhanced member items to support filtering based on search queries.
  - Updated member list items to utilize search queries for improved filtering.

- **Improvements**
  - Enhanced member list with dynamic search capabilities.
  - Implemented state management for search queries across components.

- **Technical Updates**
  - Updated GraphQL queries to support search filtering.
  - Modified type definitions to clarify interface properties.

### DIFF
--- a/packages/components/modules/profiles/Members/index.tsx
+++ b/packages/components/modules/profiles/Members/index.tsx
@@ -1,7 +1,7 @@
-import { FC, Suspense } from 'react'
+  import { FC, Suspense, useState } from 'react'
 
 import { useCurrentProfile } from '@baseapp-frontend/authentication'
-import { LoadingState as DefaultLoadingState } from '@baseapp-frontend/design-system'
+import { LoadingState as DefaultLoadingState, Searchbar } from '@baseapp-frontend/design-system'
 
 import { Typography } from '@mui/material'
 import { useLazyLoadQuery } from 'react-relay'
@@ -18,6 +18,7 @@ const Members: FC<UserMembersProps> = ({
   LoadingState,
   LoadingStateProps,
   membersContainerHeight,
+  searchQuery,
 }) => {
   const { currentProfile } = useCurrentProfile()
 
@@ -35,6 +36,7 @@ const Members: FC<UserMembersProps> = ({
       LoadingState={LoadingState}
       LoadingStateProps={LoadingStateProps}
       membersContainerHeight={membersContainerHeight}
+      searchQuery={searchQuery}
     />
   )
 }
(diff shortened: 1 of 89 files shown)


packages/components/modules/profiles/Members/index.tsx
```diff
@@ -47,23 +49,36 @@ const MembersSuspended: FC<UserMembersSuspendedProps> = ({
...
+  const [searchQuery, setSearchQuery] = useState<string>('')
+  return (
+    <>
     <Typography component="h4" variant="h4" mb={1}>
       {title}
     </Typography>
     <Typography component="p" variant="body2" color="text.secondary" mb={4}>
       {subtitle}
     </Typography>
+    <Searchbar
```
anicioalexandre: I think the `Searchbar` could be inside the `MembersList` instead since it directly used to control the search being done there.

Would be nice to use `useTransition` to control the UI loading (you can show a loading state by passing `isPending={isPending}` to  the `Searchbar`) and to use useForm on this example since it connects well with `Searchbar`, you can base on this [example](https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/messages/ChatRoomsList/index.tsx#L38)

anicioalexandre (changes requested):

packages/components/modules/profiles/Members/index.tsx
luisguareschi: @anicioalexandre Fixed!

anicioalexandre (approved):

priscilladeroode (approved):

PR #185, opened 2025-01-15
BA-2097: leave chatroom as admin

pt-tsl:
**Acceptance Criteria** 

- Given I click on the "Leave Group" button in the 3 dots menu, when the action is triggered, then I should see a confirmation prompt asking if I am sure I want to leave the group.

- Given I am the sole admin in the group chat, then the confirmation dialog should inform me prompt me to select a new admin before leaving and informing me if I don't, then the most senior member will be the new admin.

- Given there are other admins besides me in the group then only display a the same confirmation dialog a normal member would have when leaving the group chat.

- Given I confirm that I want to leave the group, then I should no longer see the group in my chat list and stop receiving updates of messages on that group.

We are going to do a hard delete for that Group Chat for that user.

- Given I've left the chat room, then make the most senior member in the group the new admin.


Ronan-Fernandes (approved):

packages/components/modules/messages/__shared__/LeaveGroupDialog/constants.ts
```diff
@@ -0,0 +1,41 @@
+export const LEAVE_GROUP_DIALOG_TEXT_COPY_ACTION_KEYS = {
+  IS_LEAVING: 'LEAVE_GROUP_DIALOG_TITLE',
```
tsl-ps2: Is this intentional? Looks like you copied `LEAVE_..._COOPY_TYPE_KEYS` and didn't change the right side (values after `:`).

packages/components/modules/messages/__shared__/LeaveGroupDialog/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+  IS_BEING_REMOVED: 'LEAVE_GROUP_DIALOG_CONTENT',
+} as const
+
+export const LEAVE_GROUP_DIALOG_TEXT_COPY_ROLE_KEYS = {
+  ADMIN: 'LEAVE_GROUP_DIALOG_ADMIN',
```
tsl-ps2: Would it make sense to rename the keys? They seem to be used as `sole admin` (ADMIN) and `not sole admin` (MEMBER) below.

packages/components/modules/messages/__shared__/LeaveGroupDialog/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+        'You will stop receiving messages from this conversation and people will see that you left.',
+    },
+  },
+  [LEAVE_GROUP_DIALOG_TEXT_COPY_ACTION_KEYS.IS_BEING_REMOVED]: {
+    [LEAVE_GROUP_DIALOG_TEXT_COPY_ROLE_KEYS.ADMIN]: {
+      [LEAVE_GROUP_DIALOG_TEXT_COPY_TYPE_KEYS.TITLE]: 'Remove from chat?',
+      [LEAVE_GROUP_DIALOG_TEXT_COPY_TYPE_KEYS.CONTENT]:
+        'Are you sure you want to remove this person from the conversation? They will no longer be able to send or receive new messages.',
+    },
+    [LEAVE_GROUP_DIALOG_TEXT_COPY_ROLE_KEYS.MEMBER]: {
```
tsl-ps2: The name `MEMBER` might be a bit confusing here: members (non-admins) should not be able to remove other participants from a group chat.

packages/components/modules/messages/utils.ts
```diff
@@ -60,3 +67,17 @@ export const getChatRoomConnections: (
   }
   return []
 }
+
+export const useCheckIsAdmin = (participants: any) => {
+  const { currentProfile } = useCurrentProfile()
+  const me = participants?.edges?.find(
```
tsl-ps2, line 73: With your backend PR `me` should always be `participants.edges[0]`, right? But maybe it's better not to rely on that and keep this `const me = participants?.edges?.find(...)` as is?

tsl-ps2 (approved):
I left some minor comments.

anicioalexandre (approved):

PR #186, opened 2025-01-17
BA-2116 System messages

tsl-ps2:
This implements system messages for creating a group chat. (The story below is slightly different: it asks for creating such messages when someone is added to a group. However, since this functionality does not exist yet, and all messages are kind of equivalent, this PR introduces them for creating a group chat.)

**Description**

As a user, on the BaseApp Messages,I would like to See a message when a new member has joined the group, In order to be aware of any changes to the group membership.

**Acceptance Criteria
Business Rules**

- Given I am a user in a group chat, when an admin has added a new member to the group, then I should see a message communicating that "[Member Name] was added".

- The user should be able to distinguish from the conversation bubbles and this system message

- The system-generated message must adhere to the chronological order of the chat history.

  - It should appear immediately after the last message sent in the chat timeline, based on the timestamp of the event when the new member joined.
- All the members in the chat group should see this message, including the person that was added.

**Notes**
Next story make it about the other system messages:

- Removed a member

- Left the group

- Create the group chat

- Edited name

- Edited picture

- Make someone an admin

## Summary by CodeRabbit

- **New Features**
	- Introduced system messages in the messaging interface.
	- Added message type differentiation between user and system messages.
	- Enhanced message rendering with type-specific components.

- **Improvements**
	- Updated GraphQL schema to support message type classification.
	- Refined message list rendering logic.
	- Improved message grouping and display mechanisms.

- **Technical Updates**
	- Updated package version to 0.0.53.
	- Restructured message-related components and types.


packages/components/modules/messages/MessagesList/MessagesGroup/index.tsx
```diff
@@ -125,41 +126,47 @@ const MessagesGroup: FC<MessagesGroupProps> = ({
...
-            <AvatarWithPlaceholder
-              className="self-start justify-self-center"
-              width={32}
-              height={32}
-              src={message?.profile?.image?.url}
-              sx={{ border: 'none' }}
-            />
-          </Box>
-        )}
+      {message.messageType === MESSAGE_TYPE.system ? (
```
anicioalexandre, line 83: maybe extract the ternary into a function/component? Just in case we need to add more messages types in the future

anicioalexandre (approved):

packages/components/modules/messages/utils.ts
```diff
@@ -31,3 +32,27 @@ export const getParticipantCountString = (participantCount: number | null | unde
...
+}
+
+export const useParsedMessageContent = (message: MessageItemCore | null | undefined) => {
+  const { currentProfile } = useCurrentProfile()
+  if (!message) return null
+  if (message.messageType === MESSAGE_TYPE.user) return message.content
+
+  const linkedProfileName =
+    message.contentLinkedProfile?.id === currentProfile?.id
+      ? 'You'
```
anicioalexandre: what if the you comes a the end of the phrase? Like `"Ana added you."`

packages/components/modules/messages/utils.ts
```diff
@@ -31,3 +32,27 @@ export const getParticipantCountString = (participantCount: number | null | unde
...
+      : message.contentLinkedProfile?.name
+  return message.content?.replace('{content_linked_profile}', linkedProfileName ?? '')
```
anicioalexandre: lets move this to a constant

packages/components/modules/messages/utils.ts
```diff
@@ -31,3 +32,27 @@ export const getParticipantCountString = (participantCount: number | null | unde
...
+    | {
+        readonly id: string
+        readonly name: string | null | undefined
+      }
+    | null
+    | undefined
+  readonly messageType: MessageTypeOptions | '%future added value' | null | undefined
+}
+
+export const useParsedMessageContent = (message: MessageItemCore | null | undefined) => {
```
anicioalexandre: I have a feeling we could generate this message on the backend

tsl-ps2: That would work, too. Then I would write a `resolve_content` function on `BaseMessageObjectType`, which generates the message based on the id of the user requesting it.

For me this is more a 'philosophical' question whether such replacements should be done on the frontend or backend. I thought of it as a question of how the content (the abstract message that some profile id did something) is presented to the user and therefore did the replacement on the frontend. But I agree that doing it on the backend leads to fewer dependencies between back- and frontend, so that the code might be easier to understand? So I am happy to switch, if you prefer that.

tsl-ps2: Moved this to the BE.

packages/components/modules/messages/utils.ts
tsl-ps2: (Outdated, since moved to BE)

packages/components/modules/messages/utils.ts
tsl-ps2: (Outdated, since moved to BE.)

packages/components/modules/messages/MessagesList/MessagesGroup/index.tsx
anicioalexandre: @tsl-ps2 can we address this as well?

PR #200, opened 2025-01-29
BA-2169 BaseApp Activity Log - Filter by Date

mathieubouhelier:
Description

As a user, on the Baseapp Activity Log,I would like to Filter activity entries by date, In order to quickly search for activities made in specific periods of time.

 
Acceptance Criteria 
Context
The objective of this story is to implement a filter that reduces the amount of activity logs displayed in order to search for logs in a specific period of time. 

Business Rules

Given I am a user searching for specific activites, when I use the date filters, Then the system shall display all activity logs that occurred from the start date through the end date, inclusive of both dates.

The system must provide a user-friendly interface for selecting a start date and an end date for filtering the activity logs.

Make sure the component complies with the current implemented Design System

The date selection tool must prevent the user from selecting an end date that is earlier than the start date.

The filtered activity log view must include a clear indicator of the date range currently being viewed per figma

Users must have the ability to clear the selected date range to return to viewing all activity logs without applying the date filter individually

The filter must apply after a date has been set, even if it's only 1 date that has been selected

The user should not be allowed to select future dates. 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=3738-46140&t=0SETItr5R9RznbuZ-4 

DEMO:

https://www.loom.com/share/dec2d3ab87984a85acc52ccce6f20a81

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive date filtering option in the activity log, allowing users to select start and end dates.
  - Responsive experience with a modal on desktop and a swipe interface on mobile.
  - Enhanced date display with a new format and improved messaging when no matching results are found.

- **Chores**
  - Updated the application version to 1.0.12.


packages/components/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @baseapp-frontend/components
 
+## 0.0.59
+
+### Patch Changes
+
+- add filter by date
```
anicioalexandre: please update versions
also, could you be more specific on the changes, like what module you are working on and a bit more details on the changes?

packages/components/modules/__shared__/MobileDrawer/index.tsx
anicioalexandre, line 1: we already have a SwipeableDrawer component at `packages/design-system/components/web/drawers/SwipeableDrawer/index.tsx`. Do you think we can reuse it?

packages/components/modules/activity-log/DateFilterChip/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const toDate = formatLabelDate(createdTo)
+    if (createdFrom && createdTo) return `${fromDate} - ${toDate}`
+
+    if (createdFrom) return `From ${fromDate}`
+    if (createdTo) return `Until ${toDate}`
+    return 'Period'
+  }
+
+  const handleOnClickOnChip = (event: React.MouseEvent<HTMLElement>) => {
+    if (isMobile) {
```
anicioalexandre, line 42: instead of doing this condition I would do:

```
if (isMobile) return (
        <Chip
        label={renderLabel()}
        onClick={handleDrawerOpen}
        variant={hasDateSelected ? 'filled' : 'soft'}
        color="default"
      />
      <MobileDrawer open={drawerOpen} onClose={handleDrawerClose} title="Period">
        {renderDateFilterComponent(handleDrawerClose)}
      </MobileDrawer>
)

return (
        <Chip
        label={renderLabel()}
        onClick={event => setAnchorEl(event.currentTarget)}
        variant={hasDateSelected ? 'filled' : 'soft'}
        color="default"
      />
      <Menu anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleMenuClose}>
        <Box padding={2}>{renderDateFilterComponent(handleMenuClose)}</Box>
      </Menu>
)
```

so we dont render mobile and web components at the same time

anicioalexandre (changes requested):
maybe we could do it in a bug, but the log timeline looks a bit off, it should continue until we reach the next profile avatar, atm the data is getting on the way. @priscilladeroode can you create a story to double check the UI on the activity-log module?

packages/components/modules/activity-log/DateFilterChip/index.tsx
mathieubouhelier: ty @anicioalexandre very good point

packages/components/modules/activity-log/DateFilterComponent/types.ts
```diff
@@ -0,0 +1,17 @@
+import { DateTime } from 'luxon'
+
+export interface IFetchParameters {
```
anicioalexandre: could you move this to `packages/components/modules/activity-log/common/types.ts`? it's probably going to be useful for native implementation, and it's something that is being reused as a log, so we should put it in a more accessible place.

Also rename it to `IFetchParameters`, we're trying to get rid of the `I` prefix 🙏🏽

anicioalexandre (approved):

anicioalexandre (approved):

PR #213, opened 2025-02-10
BA-2215: highlight existing member

pt-tsl:
**Acceptance Criteria**
**Context**
We've already implemented the Add Member List, but if a Contact is already added to the group we have no mechanism to highlight it so that the Admin doesn't add it again. 

**Business Rules - Highlight Existing Contact on Add Member List**

- Given I am on the Add Member List, when I see a contact that is already a member of the group, then I should be able to identify that its already added and I should not be able to add it again.
  - The text format should be grayed out
  - Include "Already in this group" copy instead of their username.
  - The image is NOT grayed out
- The admin should not be able to select that contact to be added 

**Design Link:** https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=6999-397517&t=Hs9bSA9tuXgxsPGa-0 

**Approvd** 
https://app.approvd.io/projects/BA/stories/38004

### DIFF
--- a/packages/components/modules/messages/EditGroup/AddMemberCard/index.tsx
+++ b/packages/components/modules/messages/EditGroup/AddMemberCard/index.tsx
@@ -15,7 +15,8 @@ const AddMemberCard: FC<AddMemberCardProps> = ({
   profile: profileRef,
   handleAddMember,
   handleRemoveMember,
-  isMember = false,
+  isBeingAdded = false,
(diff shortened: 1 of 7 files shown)


Ronan-Fernandes (approved):
Nice PR

tsl-ps2 (approved):

PR #221, opened 2025-02-17
BA-2267: generate edit username

pt-tsl:
**Acceptance Criteria 
Context**
Even though the username field exists, right now we don't have any way for users to setup their username in BA. And they are important for the different social feature that we have.

**Business Rules - Automatic Username Creation**

- Given I have successfully sign up and my user is created, automatically create a username using the name. (no spaces)
- If the generated username already exists, append a number at the end, incrementing the number until a unique username is established.
- The username must be a minimum of 8 characters long. If the name is shorter than 8 characters, append random digits until it meets the requirement.
- Users should not contain special characters

**Business Rules - Edit Username**

- Parse slashes on username and profile page so that the user doesn't have to put a "/" on its user to be used in the paths.
- *Include Slashes on the URL path in the BE
- Validate username input to check for the rules set up for user creation
- Given the user inputs a username that is already in use, suggest 1 other username.

**Functional Requirements**

- The system must automatically generate a username upon successful signup.
- The system must check the uniqueness of the generated username against the existing usernames in the database

**Approvd**
https://app.approvd.io/silverlogic/BA/stories/39057 

## Summary by CodeRabbit

- **New Features**
  - Enhanced profile settings interface with improved input validation for URL paths.
  - Upgraded phone number input now features an intuitive country select dropdown with visual cues.
  - Introduced a new `ProfileSettingsComponent` for managing user profile settings.
  - Added a new `UsernameIcon` component for consistent branding.
  - Added a new `CountrySelect` component to facilitate country selection in phone number input.
  - New GraphQL fragments for user profiles and members, enhancing data retrieval capabilities.

- **Bug Fixes**
  - Resolved display inconsistencies by automatically removing unwanted characters from usernames in profiles and chat views.

- **Documentation**
  - Updated interactive examples and documentation in Storybook for profile settings and phone number input components.


anicioalexandre (changes requested):
can we bring the storybook into here as well? I think it got lost in this migration

Ronan-Fernandes (approved):

anicioalexandre (approved):

tsl-ps2 (approved):

PR #223, opened 2025-02-21
BA-2279: mobile settings menu

anicioalexandre:
**Acceptance Criteria** 

* Given that I am on any screen within the mobile app, when I look at the bottom navigation bar, then I should see a "Settings" option/icon.
* When I tap on the "Settings" option in the bottom navigation bar, then I should be navigated to the Settings menu page.
* Given that I am on the Settings menu, then I should see my avatar, name, and username displayed at the top.
* Given I press on my avatar, name or username, it will redirect me to my profile page.
** If the profile doesn't have a username, then center the name.
* Given that I am on the Settings menu, then I should see the following options: Profile, Notifications, Login & Security, Blocked Profiles and log out option
** For now these options will not redirect anywhere, except for the Profile Settings which is already implemented.
** The log out option should be in a separate group called "Actions", while the other 4 options should be grouped in "Settings".
* The settings page must have a back button that will direct you to the previous page you were on.
* The settings page must have  close button that will direct you to the home page.
* We will have to create another route that will let you see the bottom nav bar for the Settings menu.

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive avatar component for clickable images.
  - Added several new icon visuals to enrich the interface.
  - Launched a refreshed navigation layout with enhanced back/close actions.
  - Enabled configurable keyboard dismissal for smoother input handling.

- **Refactor**
  - Streamlined the profile settings and app bar layouts for a cleaner, more flexible UI.

- **Chores**
  - Updated package versions and dependencies across the system.


packages/design-system/components/native/avatars/ClickableAvatar/styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  StyleSheet.create({
+    container: {
+      alignItems: 'center',
+      borderColor: colors.surface.active,
+      borderRadius: 50,
+      borderWidth: 2,
+      height: height + 4,
+      justifyContent: 'center',
+      overflow: 'hidden',
+      width: width + 4,
```
Ronan-Fernandes, line 16: why the + 4 here ?

Ronan-Fernandes (approved):

packages/design-system/components/native/avatars/ClickableAvatar/styles.ts
anicioalexandre: @Ronan-Fernandes there is a border outside the avatar itself with some spacement between it

PR #230, opened 2025-02-28
BA-1744: Technical Debts

priscilladeroode:
move statics from `packages/components/modules/navigations/Header/AccountMenu/AccountPopover/_storybook_/static/profile-1.png` to `packages/components/.storybook/statics`

improve packages/components/modules/navigations/Header/AccountMenu/AccountPopover/_tests_/AccountPopover.cy.tsx tests

duplicated test descriptions (we should probably merge it into one)

replace cy.contains with findByText

instead of querying within cy.findByLabelText('List of available profiles') we could've used findByAltText using the AvatarWithPlaceholder's alt ("Profile avatar")

at packages/components/modules/navigations/Header/AccountMenu/AccountPopover/index.tsx, AddProfileMenuItem should not be children of LogoutItem since they have no logic relation. Let's render both separately (an outside container may be needed)

at packages/components/modules/profiles/ProfilePopover/ProfilesList/ProfileMenuItem/index.tsx, double check if we want to keep using readInlineData

evaluate if we can move and use more granular configs like tsconfig.eslint.json, and tsconfig.jest.json to the tsconfig package to make it available to other packages

Refactor AccountPopover component to improve readability. 

Approvd 
https://app.approvd.io/projects/BA/stories/36373 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `CurrentProfileMenu` component for enhanced user profile management.
- **Refactor**
	- Consolidated rendering logic in the `AccountPopover` component for improved clarity and structure.
	- Simplified the `LogoutItem` component by removing unnecessary wrapper elements.
	- Updated the `AccountPopoverProps` interface to streamline properties.
- **Tests**
	- Enhanced test cases for user-related elements, improving efficiency and reducing redundancy.
- **Chores**
	- Updated version number to `1.0.23` in the package.

### DIFF
--- a/packages/components/modules/navigations/web/Header/AccountMenu/AccountPopover/types.ts
+++ b/packages/components/modules/navigations/web/Header/AccountMenu/AccountPopover/types.ts
@@ -2,23 +2,14 @@ import type { FC } from 'react'
 
 import type { CSSProperties } from '@mui/material/styles/createMixins'
 
-import type {
-  AddProfileMenuItemProps,
-  ProfilesListProps,
-  SwitchProfileMenuProps,
-} from '../../../../../profiles/web'
+import type { AddProfileMenuItemProps, ProfilesListProps } from '../../../../../profiles/web'
+import { CurrentProfileMenuProps } from './CurrentProfileMenu/types'
 import type { LogoutItemProps } from './LogoutItem/types'
-import type { MenuItemsProps } from './MenuItems/types'
 
-export interface AccountPopoverProps {
+export interface AccountPopoverProps
+  extends Omit<CurrentProfileMenuProps, 'handlePopoverOnClose' | 'setOpenProfilesList'> {
   PopoverStyles?: CSSProperties
-  CurrentUser?: FC | null
-  CurrentProfile?: FC | null
-  MenuItems?: FC<MenuItemsProps> | null
-  MenuItemsProps?: Partial<MenuItemsProps>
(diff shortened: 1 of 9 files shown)


deboracosilveira (approved):

deboracosilveira (approved):

packages/components/modules/navigations/web/Header/AccountMenu/AccountPopover/types.ts
```diff
@@ -2,23 +2,14 @@ import type { FC } from 'react'
...
-  SwitchProfileMenuProps,
-} from '../../../../../profiles/web'
+import type { AddProfileMenuItemProps, ProfilesListProps } from '../../../../../profiles/web'
+import { CurrentProfileMenuProps } from './CurrentProfileMenu/types'
 import type { LogoutItemProps } from './LogoutItem/types'
-import type { MenuItemsProps } from './MenuItems/types'
 
-export interface AccountPopoverProps {
+export interface AccountPopoverProps
+  extends Omit<CurrentProfileMenuProps, 'handlePopoverOnClose' | 'setOpenProfilesList'> {
```
anicioalexandre, line 10: @priscilladeroode I think we should've `Pick` the ones we want instead of `Omit`, so we don't need to worry if we ever add new props to `CurrentProfileMenu` that should not be part o `AccountPopoverProps`

PR #237, opened 2025-03-10
BA-2351 user avatar icon not loading immediately

matheusysd:
- @baseapp-frontend/components package update - v 1.0.17
   - Fix image path handling for profile images
   - Updated dependencies
      - @baseapp-frontend/authentication@4.1.8

- @baseapp-frontend/authentication package update - v 4.1.8
   - Fix image path handling for profile images

DEMO: https://www.loom.com/share/5fdb6987a463459cbea51a80bbb5ddf9?sid=740b0a30-621d-4b90-b1e0-d0e487a9f44b

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced profile image display by ensuring image URLs are correctly handled whether they are external or local.
  
- **Chores**
	- Updated package versions to reflect the latest improvements and dependency upgrades.

### DIFF
--- a/packages/authentication/modules/access/useLogin/index.ts
+++ b/packages/authentication/modules/access/useLogin/index.ts
@@ -61,7 +61,12 @@ const useLogin = ({
       if (user) {
         // TODO: handle the absolute image path on the backend
         const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL?.replace('/v1', '')
-        const absoluteImagePath = user?.profile?.image ? `${baseUrl}${user.profile.image}` : null
+        let absoluteImagePath = null
+        if (user?.profile?.image) {
+          absoluteImagePath = user.profile.image.startsWith('http')
+            ? user.profile.image
+            : `${baseUrl}${user.profile.image}`
+        }
         setCurrentProfile({
           ...user.profile,
           image: absoluteImagePath,
--- a/packages/authentication/CHANGELOG.md
+++ b/packages/authentication/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @baseapp-frontend/authentication
 
+## 4.1.8
+
+### Patch Changes
+
+- Fix image path handling for profile images
+
 ## 4.1.7
 
 ### Patch Changes
--- a/packages/authentication/package.json
+++ b/packages/authentication/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@baseapp-frontend/authentication",
   "description": "Authentication modules.",
-  "version": "4.1.7",
+  "version": "4.1.8",
   "main": "./index.ts",
   "types": "dist/index.d.ts",
   "sideEffects": false,
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @baseapp-frontend/components
 
+## 1.0.17
+
+### Patch Changes
+
+- Fix image path handling for profile images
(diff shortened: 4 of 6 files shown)


packages/authentication/modules/access/useLogin/index.ts
```diff
@@ -61,7 +61,12 @@ const useLogin = ({
       if (user) {
         // TODO: handle the absolute image path on the backend
```
deboracosilveira, line 62: what about this TODO? why not handling it on BE?

matheusysd: To be honest, I didn't know if it was planned to solve this in BE. If I'm not mistaken @tsl-ps2 created this TODO, but when we discussed how we could fix this bug while keeping both behaviors the solution we came up with was this way I did it
I will add him here so he can share what he thought
Also I didn't realize we could fix this on BE, I was about to ask how we could do it then I though that we can append base url when sending image path to FE
cc: @tsl-ps2

nossila: I think its better to never assume the `baseUrl` for the image in the frontend, and always just use whatever the backend returns. But I'm fine with covering this here... but if was a backend concern then our code would look like:

```const absoluteImagePath = user?.profile?.image```

making the const unnecessary actually, and no if statement is necessary as well

seems like we can remove that todo in my opinion as well

tsl-ps2: To clarify: The issue is that sometimes (when running locally and probably also on baseapp staging/production since the issue did not appear there) the paths returned by the backend are local paths, but for RPP the returned path is global (probably has to do with some server settings, I don't know). On the backend, we usually use the `request` object to convert all paths to global paths. The issue here is that the base profile information is serialized by the `get_token` method of the `BaseJWTLoginSerializer`; this method is implemented by the simple JWT package as a `@classmethod` and overwritten by a custom method. However, the request object is not available in this method and therefore the conversion to absolute paths does not work automatically. In the backend PR corresponding to the changes I had made in December, I had removed the `@classmethod` decorator, so that the request became available in `get_token` through `self.context`, see this comment: https://github.com/silverlogic/baseapp-backend/pull/191#discussion_r1867799665. We discussed this in Bravo and came to the conclusion that overwriting a classmethod with a non-classmethod is not safe and could potentially break things if the way the simple JWT package works internally changes. Therefore I reverted those BE changes and did the conversion on the frontend instead.

tsl-ps2: Using just `const absoluteImagePath = user?.profile?.image` without making other changes would break loading the profile image when running the code locally, or whenever the backend returns local paths (of the form "/media/..."). Without prefixing, such files with local paths are resolved as `localhost:3000/media/...` but we need them to be resolved as `localhost:8000/media/...` (when running locally with standard ports)

tsl-ps2 (approved):

packages/authentication/modules/access/useLogin/index.ts
nossila: This looks like a issue with the current JWT setup, but since `baseapp_auth` is being rewriting maybe it will get fixed, lets see. I'm fine with this frontend implementation.

nossila: Thanks for clarifying @tsl-ps2

nossila (approved):

anicioalexandre (approved):

PR #239, opened 2025-03-12
BA-2364: change password

pt-tsl:
**Acceptance Criteria**

- Given that I am on the Login & Security settings page, when I tap on the "Change Password" option, then the section should expand to reveal the "Old Password," "New Password," and "Confirm New Password" input fields.
- When all fields are filled and the New Password matches the Confirm New Password, then the "Change Password" button should be enabled.
- Display error alerts for:
  - When the old password is incorrect
  - When the new password doesn't follow our current password validation requirements (symbols, numbers, caps, etc.)
  - When the new password doesn't match with confirm new password.
- When I tap on the "Change Password" button, then the system should validate my old password and update my password if the validation is successful.
- Given that I am in the process of changing my password, when I tap the "Cancel" button, then the section should collapse, any entered data should be cleared, and I should return to the default Login & Security settings view.
- By default the input appear in a hidden state, but the user should be able to unhide by clicking the "eye" icon on the right side of the input box.
- This feature is already in the web template, but we need to move the form and the form mutation to the authentication package.

**Guardrails**
Although the flow in Figma shows a modal asking the user if they want to end all current sessions, we have to check if that functionality was already built by PORG.
So for now just update the password without showing the modal. 

**Design Link:** https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=6046-17914&t=ZYZocuoVLM11VrtX-0 

**Approvd** 
https://app.approvd.io/silverlogic/BA/stories/39018 

## Summary by CodeRabbit

- **New Features**
	- Rolled out an enhanced password change experience for smoother account management.
	- Introduced a new alert icon for improved visual notifications.
	- Added a scrollable view component with built-in keyboard handling for better content navigation.
- **Enhancements**
	- Improved error display in text inputs to prevent overflow and ensure clarity.
	- Updated button styling for more consistent and legible text presentation.
- **Updates**
	- Upgraded key packages across authentication, components, and design system for a more integrated and reliable user experience.


packages/design-system/components/native/views/ScrollView/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { createStyles } from './styles'
+import type { ScrollViewProps } from './types'
+
+const ScrollView: FC<ScrollViewProps> = ({ style, avoidKeyboard = true, children, ...props }) => {
+  const styles = createStyles()
+
+  if (avoidKeyboard) {
+    return (
+      <KeyboardAvoidingView behavior="padding" style={{ flex: 1 }}>
+        <NativeScrollView style={[styles.container, style]}>{children}</NativeScrollView>
```
anicioalexandre: let's keep passing the `...props` to this native view

packages/design-system/components/native/views/ScrollView/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { createStyles } from './styles'
+import type { ScrollViewProps } from './types'
+
+const ScrollView: FC<ScrollViewProps> = ({ style, avoidKeyboard = true, children, ...props }) => {
+  const styles = createStyles()
+
+  if (avoidKeyboard) {
+    return (
+      <KeyboardAvoidingView behavior="padding" style={{ flex: 1 }}>
```
anicioalexandre: lets move this to a style sheet

anicioalexandre (approved):

packages/design-system/components/native/inputs/TextInput/index.tsx
```diff
@@ -32,7 +39,9 @@ const TextInput: FC<TextInputProps> = (props) => {
         {...props}
       />
       {helperText && !disabled && (
-        <View style={styles.errorContainer}>
+        // Had to do this adjustment to the error container width because the error text was overflowing the container
+        // The 12px subtraction is to account for the padding of the error container
+        <View style={[styles.errorContainer, { width: errorContainerWidth - 12 }]}>
```
tsl-ps2, line 44: Some advanced nitpicking: Just looking at the code, it seems to me that `errorContainerWidth - 12` is the width of the container, so you could also subtract 12 in the `onLayout` method (if you think this makes sense).

tsl-ps2 (approved):

packages/design-system/components/native/inputs/TextInput/index.tsx
pt-tsl: Not entirely sure why, but doing so doesn't work.

anicioalexandre (approved):

PR #243, opened 2025-03-21
BA-2392 RN Notifications: Divide List by Read

tsl-ps2:
Description

As a user, on the BaseApp React Native,I would like to see dividers in my notification list based on how old the notifications are, In order to distinguish between recent and older notifications.

 
Acceptance Criteria 

- Given I have multiple notifications, when I open the notifications page, then the notifications should be grouped by Read/Unread and separated by a divider.

- The divider will have the "Read" Label

- The reverse chronological sort should be maintained inside each group. (Most recent first)

- Given a notification is mark as read, then it should go to the read group live.

  - This is for the case when we click on "Mark all as read button", where we don't redirect to another page.

- If everything is read, then the divider should be hidden. 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/38686 

## Summary by CodeRabbit

- **New Features**
  - Introduced a visual divider in the notifications list to enhance readability.
  - Enabled an automatic refresh for notifications after marking them as read.
  - Optimized the notifications fetching strategy for a smoother and more responsive experience.
  - Added a new `Divider` component to improve layout structure within the notifications list.
- **Bug Fixes**
  - Improved organization of notifications by categorizing them into unread and read sections.

### DIFF
--- a/packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
+++ b/packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
@@ -10,6 +10,7 @@ import { NotificationItemProps } from './types'
 const NotificationItem: FC<NotificationItemProps> = ({
   notification: notificationRef,
   NotificationItemRenderer = DefaultNotificationItemRenderer,
+  refetch,
 }) => {
   const notification = useFragment(NotificationItemFragment, notificationRef)
 
@@ -18,6 +19,7 @@ const NotificationItem: FC<NotificationItemProps> = ({
   const markAsRead = () => {
     if (notification.unread) {
       commitMutation({
+        onCompleted: refetch,
         variables: {
           input: {
             read: true,
--- a/packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
+++ b/packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
@@ -19,6 +19,8 @@ const NotificationItem: FC<NotificationItemProps> = ({
   const markAsRead = () => {
     if (notification.unread) {
       commitMutation({
+        // TODO: check whether this is needed (clicking on a notification should navigate to the underlying comment,
+        // so the screen will not remain visible). If needed, check whether we can use optimistic updates
         onCompleted: refetch,
         variables: {
           input: {
--- a/packages/components/modules/notifications/native/NotificationsList/index.tsx
+++ b/packages/components/modules/notifications/native/NotificationsList/index.tsx
@@ -16,6 +16,7 @@ import {
   useNotificationsSubscription,
 } from '../../common'
 import { NotificationsNode } from '../../common/types'
+import Divider from './Divider'
 import DefaultEmptyState from './EmptyState'
 import MarkAllAsReadButton from './MarkAllAsReadButton'
 import DefaultNotificationItem from './NotificationItem'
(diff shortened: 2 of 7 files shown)


packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
```diff
@@ -18,6 +19,7 @@ const NotificationItem: FC<NotificationItemProps> = ({
   const markAsRead = () => {
     if (notification.unread) {
       commitMutation({
+        onCompleted: refetch,
```
anicioalexandre, line 24: let's add a TODO: replace this with optimistic updates
do we have time on this story to do it?
cc: @Ronan-Fernandes

packages/components/modules/notifications/native/NotificationsList/index.tsx
```diff
@@ -48,19 +49,26 @@ const NotificationsList: FC<NotificationsListProps> = ({
...
 
-  const renderNotificationItem = (notification: any) => {
+  const renderNotificationItem = (notification: any, index: number) => {
     if (!notification) return null
-
+    let divider = null
+    if (!notification.unread && index > 0 && notifications[index - 1]?.unread) {
+      divider = <Divider />
+    }
     // TODO add notifications divider and unread/Read notifications as per design
```
anicioalexandre: lets remove the TODO :D

packages/components/modules/notifications/native/NotificationsList/index.tsx
```diff
@@ -69,7 +77,10 @@ const NotificationsList: FC<NotificationsListProps> = ({
 
     return (
       <View>
-        <FlatList data={notifications} renderItem={({ item }) => renderNotificationItem(item)} />
+        <FlatList
+          data={notifications}
+          renderItem={({ item, index }) => renderNotificationItem(item, index)}
```
anicioalexandre: @Ronan-Fernandes could you coordinate this PR with yours (infinite scroller)? we might have conflicts here

anicioalexandre (approved):

packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
tsl-ps2: @anicioalexandre: Not sure whether we need optimistic updates here. I think FS said clicking on a notification should navigate to that page (but this is not implemented yet), in which case the user would not see the optimistic update anyways. I refetched this so that the new data is already there when the user returns to the notifications page. Maybe we could even delay the refetch by a second, so that it does not slow down loading of the new page?

packages/components/modules/notifications/native/NotificationsList/index.tsx
Ronan-Fernandes: yeah it is already showing the conflict here since we just merged the infinite scroller PR.

@tsl-ps2 we need to rebase this PR with master and adapt this code to work with the infinite scroller.
since you just added the index there I think this wont be a lot of work

packages/components/modules/notifications/native/NotificationsList/NotificationItem/index.tsx
Ronan-Fernandes: @tsl-ps2 we will already do the optimistic updates on the mark all as read, so maybe we do the same here as well.
could even be the same story.
I don't know if @fs already did this story, but we could talk about it with him and add this piece there.

Ronan-Fernandes (approved):

anicioalexandre (approved):

PR #246, opened 2025-03-27
BA-2399 [RN] Comment Creation

tsl-ps2:
**Description**

**Acceptance Criteria** 
**Basic Comment Creation**
Since we wont have flat pages in React native and the Content feeds is not ready yet. We need a place to "host" comments in order to test them.
So we will create a comment route on the RN application to do so.

**Business Rules**

- Create a "Comment" route where we will be able to test the comment component and its functionalities

- Recreate the comment component for React Native

- Users should be able to create a comment via an input box. 

**Comment Component Behavior**

- Given a user clicks on the input box, the mobile keyboard should slide up in order for the user to type in the content.

  - The input box will slide up with the keyboard, and it should remain visible.

- Given a user presses on the Enter Key in the keyboard, it should create a line break in the content.

- Given a user has finished creating a comment, when they press the "send button", then the comment should be created.

- Given the comment has been created, then the new comment should appear at the top of the comment list.

**Reaction Component Behavior**

- Each comment should display the user’s avatar, name, and timestamp

- Users should be able to react to the comment by pression the heart icon.

- Given the user presses the Reaction Icon, the count should increase by 1

- Given the same user presses the reaction icon again after having already reacted to the comment., the count should decrease by 1.

**Guardrails**

- Threads wont be implemented for now.

- Load only the first 5 comments. Pagination will be handled later 

- Design Link: https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=942-18665&t=jbYKzqIHwPMDjdiS-0 

**Notes** 

- create a temporary model and inherit the Commentable model

- we can reuse the reply logic (e.g. `useCommentReply`) and the queries on the web 

**Approvd** 
[https://app.approvd.io/silverlogic/BA/stories/39781](https://app.approvd.io/silverlogic/BA/stories/39781) 

## Summary by CodeRabbit

* **New Features**
  * Introduced a complete set of native comment components, including comment creation, display, reaction, and reply functionality.
  * Added customizable social input components with keyboard-aware drawers and enhanced input experiences.
  * Added new design system icons: Emoji, Favorite, Favorite Selected, Reply, and Send Message.
  * Added a timestamp component for displaying human-readable dates.
  * Provided context and hooks for managing social text input state.

* **Enhancements**
  * Streamlined exports for shared and native components, making types and components more accessible.
  * Improved modularity and customization options for comment and input components.

* **Chores**
  * Added `react-native-safe-area-context` as a peer dependency.


packages/components/modules/__shared__/native/SocialInputDrawer/styles.ts
```diff
@@ -0,0 +1,12 @@
+import { StyleSheet } from 'react-native'
+
+export const createStyles = () =>
+  StyleSheet.create({
+    bottomSheetContainer: {
+      paddingHorizontal: 16,
+      paddingVertical: 12,
```
anicioalexandre: we need to add a background here,on dark mode it just shows a white background
lets use theme.colors.surface.background

packages/components/modules/__shared__/native/Timestamp/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+const Timestamp: FC<TimestampProps> = ({ date }) => {
+  const dateTime = DateTime.fromISO(date)
+  const now = DateTime.now()
+  const diffDays = now.startOf('day').diff(dateTime.startOf('day'), 'days').days
+
+  let text
+  if (diffDays === 0) text = 'Today'
+  else if (diffDays === 1) text = 'Yesterday'
+  else if (diffDays <= 7) text = `${diffDays} days ago`
+  else text = formatDate(dateTime, { toFormat: DATE_FORMAT[2] })
```
anicioalexandre, line 21: I wonder if we could reuse formatDateWithDiffNow or formatRelativeTime? maybe we can review that with the design team

packages/components/modules/__shared__/native/SocialInputDrawer/styles.ts
anicioalexandre: After we change the background, we have another issue with the outer layer:
image

packages/design-system/components/native/inputs/SocialTextInput/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        ref={ref}
+        render={renderTextInput}
+        style={styles.input}
+        textColor={theme.colors.object.high}
+        {...props}
+        onBlur={handleBlur}
+        onFocus={handleFocus}
+      />
+    ) : (
+      <NativeTextInput
```
anicioalexandre: We might need to handle the input color for native too
image

anicioalexandre (changes requested):

packages/components/modules/__shared__/native/SocialInputDrawer/styles.ts
tsl-ps2: Fixed

packages/design-system/components/native/inputs/SocialTextInput/index.tsx
tsl-ps2: Fixed

packages/components/modules/__shared__/native/Timestamp/index.tsx
tsl-ps2: Very much in favor of this one since reusing existing functions also means that the dates/times we are using are more uniform (and not different for comments, messages, etc). But ultimately the design team will have to decide that.

anicioalexandre: @tsl-ps2 could you add a TODO here please? so we can address that later?

packages/components/modules/comments/native/CommentCreate/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+            form.reset()
+            if (ref && 'current' in ref) ref.current?.blur()
+          }
+        },
+        onError: console.error,
+      })
+    }
+
+    const styles = createStyles()
+    const { textHeight, isFocused } = useSocialTextInput()
```
anicioalexandre: I think we could try to simplify this component and avoid these stores with providers as much as we can
for now, `textHeight`and `isFocused` logic could be moved to the SocialInputDrawer, something like:
```
        const body = form.watch('body')
        const hasBody = body !== ''
        
        return (<SocialInputDrawer
          form={form}
          isLoading={isMutationInFlight}
          keyboardHeight={keyboardHeight}
          hasBody={hasBody} // pass this prop to calculate showHandle internally
          keyboardHeight={keyboardHeight} // pass this prop to change the internal View styling
          ref={ref}
          style={drawerStyle}
          submit={onSubmit}
          {...SocialInputDrawerProps}
        />
      </>)
```
then we move `ScrollView` inside `SocialInputDrawer`, would be no need for zustand for now

packages/components/modules/comments/native/CommentCreate/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const showHandle = isFocused || body !== ''
+
+    const [keyboardHeight, setKeyboardHeight] = useState(0)
+    const insets = useSafeAreaInsets()
+    useEffect(() => {
```
anicioalexandre: I have a feeling this logic should be better placed at `SocialInputDrawer`, no?
Let's keep in mind we will have a `MessageCreate` component in the future that should leverage the most it can from `SocialInputDrawer`

packages/components/package.json
```diff
@@ -115,7 +115,8 @@
     "@baseapp-frontend/graphql": "workspace:*",
     "@baseapp-frontend/utils": "workspace:*",
     "react": "catalog:react19",
-    "react-dom": "catalog:react19"
+    "react-dom": "catalog:react19",
+    "react-native-safe-area-context": "5.4.1"
```
anicioalexandre: lets add this to the react-native catalog and use it (ideally also for the template PR)

packages/components/modules/__shared__/native/SocialInputBox/index.tsx
anicioalexandre, line 1: why is this called `SocialInputBox`? The naming is a bit confusing, since that's the input itself with actions. Maybe just SocialInput?

packages/design-system/components/native/inputs/SocialTextInput/styles.ts
```diff
@@ -0,0 +1,65 @@
...
+    container: {
+      display: 'flex',
+      flexDirection: 'column',
+      alignItems: 'flex-start',
+      gap: 8,
+      padding: 12,
+      borderRadius: 8,
+      borderWidth: 1,
+      borderStyle: 'solid',
+      borderColor: isFocused ? theme.colors.object.high : theme.colors.surface.border,
```
anicioalexandre, line 27: I think if we had reused the BA TextInput with multiline prop would've been easier since it has a bunch of styling like this one already out of the box

packages/components/modules/__shared__/native/Timestamp/index.tsx
tsl-ps2: Done!

packages/components/modules/__shared__/native/SocialInputBox/index.tsx
tsl-ps2: I thought about it as "Text Box" or "Input Box", hence the name (wanted to put a word like box or field there to distinguish it from the drawer), but I can see that this is confusing. Changed to `SocialInput`.

packages/design-system/components/native/inputs/SocialTextInput/styles.ts
tsl-ps2: At the time I created the PR this was not possible: The BA `TextInput` uses `react-native-paper`'s input field, the `SocialTextInput` uses the plain `react-native` input. The `react-native-paper` input with non-fixed height used some constants that could not be overwritten (I think for the minimal height or something similar, I don't quite remember), but I needed to update these to comply with our designs. It might be that the package was updated since then, so maybe it's possible now. Will check quickly.

tsl-ps2: I'm not sure how to use `react-native-paper`'s Text input: [The code](https://github.com/callstack/react-native-paper/blob/v5.14.5/src/components/TextInput/TextInputOutlined.tsx) contains this:
```
<View
        style={[
              styles.labelContainer,
              {
                     paddingTop,
                     minHeight,
              },
        ]}
>
```
where `minHeight` is either the height if passed as prop or computed from their constants. But their TextInputs are really high, so when computed from their constants, the value is too big. The only way around this is that you need to pass a `height` explicitly, which means you need to watch for text changes and control it. I could make it work on Android, but not on iOS.

tsl-ps2: So it's unclear to me how to reuse BA TextInput code (unless we switch from `react-native-paper` to `react-native` in there), and I would leave it as is for now.

packages/components/modules/comments/native/CommentCreate/index.tsx
tsl-ps2: Yes and no. It was indeed on the drawer before, but I also needed the keyboard data on the `CommentCreate` component (or whichever component has access to the scroll view behind the `BottomSheet`). I now refactored the `SocialInputDrawer` as providing an object `{Drawer, Placeholder, useKeyboardHeight}`, then when you implement it, you do something like
```
const keyboardHeight = SocialInputDrawer.useKeyboardHeight()
return (
   <>
        <ScrollView>
             // Your content here
             <SocialInputDrawer.Placeholder keyboardHeight={keyboardHeight} ... />
        </ScrollView>
        <SocialInputDrawer.Drawer keyboardHeight={keyboardHeight} ... />
   </>
)
```
I hope this makes it clearer what is going on.

packages/components/modules/comments/native/CommentCreate/index.tsx
tsl-ps2: I am very hesitant to move `ScrollView` inside `SocialInputDrawer`. The drawer doesn't really need to know what is happening behind it and wrap this content in a `ScrollView`. For example, if we reuse the drawer for messages, we probably want to add some custom logic to this `ScrollView` (e.g. when scrolled up, load more messages). So ideally it would be directly accessible, not "hidden" inside the drawer component.

I also do not want to "lift the `textHeight` and `isFocused` state" to the `SocialInputDrawer`, because `SocialInput` (just the text box) is a component that I think should work on its own (it's much simpler since it doesn't require a bottom sheet, still provides a TextInput that can be focused and defocused by clicking on it and on the background, it's just lacking the second option to focus/defocus using the drawer handle).

However, I can understand that we want to avoid using `zustand` if possible. I now switched to another solution with the state still living on the `TextInput`, but a `onTextHeightChange` and `onFocusChange` handler that basically allows the `SocialInputDrawer` to access these variables (at the cost of some small overhead: there will be one "copy" of these two variables).

Let me know what you think.

packages/components/modules/__shared__/native/SocialInputDrawer/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+        ref={ref}
+        showHandle={showHandle}
+        submit={onSubmit}
+      />
+    </>
+  )
+}
+*/
+
+const SocialInputDrawer: SocialInputDrawerType = {
```
tsl-ps2, line 62: @anicioalexandre This is basically how I refactored the Drawer, check the example above and let me know what you think.

anicioalexandre (approved):

packages/components/modules/__shared__/native/SocialInputDrawer/index.tsx
anicioalexandre: I liked it @tsl-ps2

packages/components/modules/comments/native/CommentCreate/index.tsx
anicioalexandre: looks good!

lokmanliton:
Working fine now, thanks.

ficast (approved):

nossila (approved):

PR #257, opened 2025-04-23
BA-2455-fe-customizable-styles-for-main-page-area

hpmoreira05:
Acceptance Criteria 

As a developer using the baseapp-frontend,

I want to customize the main area’s styles via props,

So that I can adapt the layout and design to match my project's branding and needs.
 

Scenario: Apply custom styles to the main area via props

Given I am using the baseapp-frontend in my project

And I pass custom style props to the MainArea component

When the page renders

Then the main area should reflect the styles defined in the passed props

And the styles should override the default styles provided by the template
 

Scenario: Use default styles when no custom props are provided

Given I am using the baseapp-frontend

And I do not pass any style props to the MainArea component

When the page renders

Then the main area should display using the default styles provided by the template
 

Scenario: Partially override styles with selective props

Given I pass only some style props to the MainArea component

When the page renders

Then the main area should use the provided custom styles where specified

And it should fall back to default styles for properties that were not passed
 

Scenario: Update styles dynamically via props

Given I have rendered the MainArea with a set of style props

When I update the props at runtime (e.g., via state change or theme switch)

Then the main area should re-render with the new styles

And the changes should apply without a full page reload
 

Scenario: Prevent style conflicts between default and custom styles

Given I am customizing styles via props

When the props target the same CSS properties as defaults

Then the custom props should take precedence without causing style conflicts or duplication 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/40783 

## Summary by CodeRabbit

- **New Features**
  - Added support for passing additional properties to navigation layout containers for enhanced customization.
  - Introduced the ability to override the main container component within navigation layouts.

- **Style**
  - Improved handling to prevent internal props from being forwarded to DOM elements, ensuring cleaner rendered HTML.

### DIFF
--- a/packages/components/modules/navigations/web/NavigationLayout/index.tsx
+++ b/packages/components/modules/navigations/web/NavigationLayout/index.tsx
@@ -24,6 +24,7 @@ const NavigationLayout: FC<NavigationLayoutProps> = ({
   AccountMenuProps,
   ToolbarProps,
   children,
+  BodyProps,
 }) => {
   const nav = useBoolean()
 
@@ -45,7 +46,9 @@ const NavigationLayout: FC<NavigationLayoutProps> = ({
         >
           <NavCentered navData={navData} openNav={nav.value} onCloseNav={nav.onFalse} />
         </Header>
-        <MainContainer isNavCentered>{children}</MainContainer>
+        <MainContainer isNavCentered BodyProps={BodyProps}>
+          {children}
+        </MainContainer>
       </>
     )
   }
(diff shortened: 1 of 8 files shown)


packages/components/modules/navigations/web/NavigationLayout/index.tsx
```diff
@@ -24,6 +24,7 @@ const NavigationLayout: FC<NavigationLayoutProps> = ({
   AccountMenuProps,
   ToolbarProps,
   children,
+  BodyProps,
```
anicioalexandre: I think we could make this a bit more flexible and add `MainContainerProps` instead of a specific prop and also allow the main container to be overridden, something like:

```
import DefaultMainContainer from './MainContainer'

const NavigationLayout: FC<NavigationLayoutProps> = ({
  ...
  MainContainer = DefaultMainContainer,
  MainContainerProps = {},
}) => {
  ...
  <MainContainer isNavCentered {...MainContainerProps}>
          {children}
  </MainContainer>
```

anicioalexandre (approved):

PR #265, opened 2025-06-25
feature/BA-2525-payments-transaction-history

lbjunq:
As a User, on the BaseApp Profile Page,I would like to View a detailed history of my transactions and access receipts for my subscription payments, In order to Keep track of my subscription expenses, verify payments, and maintain accurate financial records.

 
Acceptance Criteria 

Given the user navigates to the "Transaction History" section within settings, the user views a list of all historical payments, sorted by date with the most recent payment listed first.

Use the Material UI table component

Ensure the transaction history is responsive and provides a consistent user experience on desktop, tablet, and mobile devices. 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7279-192622&t=pkTsTldiYoAVvZuw-4 

Approvd 
https://app.approvd.io/projects/BA/stories/39485 

Demo
https://www.loom.com/share/d293f0dec8a045a7aa1167a45d56df1d


rodean1 (approved):

packages/components/modules/payments/web/InvoiceListTable/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+                <TableRow {...rowProps}>
+                  <TableCell colSpan={5} sx={{ p: 0, paddingX: 2 }}>
+                    <Typography variant="body2" sx={{ fontWeight: 'bold' }}>
+                      No receipts found
+                    </Typography>
+                  </TableCell>
+                </TableRow>
+              ) : (
+                receipts.map((row: Invoice) => {
+                  const color = getStatusColor(row.status)
```
ds-tsl: lets extract this to a component itself

packages/components/modules/payments/web/InvoiceListTable/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+                            window.open(row.hostedInvoiceUrl, '_blank')
+                          }}
+                          disabled={!row.hostedInvoiceUrl}
+                        >
+                          <ChevronIcon position="right" />
+                        </IconButton>
+                      </TableCell>
+                    </TableRow>
+                  ) : (
+                    <TableRow key={row.id} {...rowProps}>
```
ds-tsl: lets also extract this to its own component

packages/components/modules/payments/web/hooks/useStripeHook.tsx
```diff
@@ -136,6 +137,12 @@ const useStripeHook = () => {
       mutationKey: ['useCreationSubscription'],
     })
 
+  const useListInvoices = ({ page = 1 }) =>
+    useQuery({
+      queryKey: ['useListInvoices'],
```
ds-tsl: we should also use a const for this querykey like the one above

packages/components/modules/payments/web/types.ts
```diff
@@ -51,9 +51,26 @@ export interface CreateSubscriptionOptions {
...
-  remote_customer_id: string
+  customer_id: string
   price_id: string
   payment_method_id?: string
   allow_incomplete?: boolean
   billing_details?: CreateSubscriptionOptions['billingDetails']
 }
+
+export type UpdatePaymentMethodRequestBody = Partial<PaymentMethod> & {
+  customer_id: string
```
ds-tsl: lets only use camel case on types

packages/components/modules/payments/web/types.ts
```diff
@@ -51,9 +51,26 @@ export interface CreateSubscriptionOptions {
...
+  default_payment_method_id?: string
+}
+
+export interface Invoice {
+  id: string
+  amountDue: number
+  status: string
+  webhooksDeliveredAt: string
+  hostedInvoiceUrl: string
+  lines: {
```
ds-tsl: we should avoid defining nested structures inline like this. It would be better to extract it into its own interface and reference it here.

packages/components/modules/payments/web/services/stripe.ts
```diff
@@ -55,7 +57,7 @@ class StripeApi {
     billingDetails,
   }: CreateSubscriptionOptions): Promise<Subscription> => {
     const requestBody: SubscriptionRequestBody = {
-      remote_customer_id: customerId,
+      customer_id: customerId,
```
ds-tsl: isn’t our axios instance already configured to handle case conversion? If so, we should stick to using camel case in our code

ds-tsl (changes requested):

packages/components/modules/payments/web/InvoiceListTable/components/InvoiceItemWrapper.tsx
```diff
@@ -0,0 +1,39 @@
...
+const InvoiceItemWrapper = ({ row, rowProps, cellProps, smDown }: InvoiceItemWrapperProps) => {
+  const amountDue = (row.amountDue / 100).toFixed(2) ?? ''
+  const formattedDate =
+    formatDateFromApi(row.webhooksDeliveredAt, { toFormat: DATE_FORMAT[1] }) ?? ''
+
+  const getStatusColor = (status: string) => {
+    if (status === 'paid') return 'success'
+    if (status === 'failed') return 'error'
+    return 'warning'
+  }
```
deboracosilveira, line 26: lets use enums here instead of the strings directly

packages/components/modules/payments/web/InvoiceListTable/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+                      count={(data?.count ?? 0) / 10}
+                      page={page}
+                      onChange={(_, value) => setPage(value)}
+                    />
+                  </Box>
+                </TableCell>
+              </TableRow>
+            </TableFooter>
+          </Table>
+        </TableContainer>
```
deboracosilveira, line 85: lets break it down into small components like header, no results, results and footer

deboracosilveira (approved):

packages/components/modules/payments/web/InvoiceListTable/components/InvoiceItemWrapper.tsx
anicioalexandre: agreed, but lets use constants instead of enums, we're trying to avoid enums since it come with some bugs. We can follow this pattern:
* constant: https://github.com/silverlogic/baseapp-frontend/blob/c29cb34c59309b44a110cdb8baa35e6871c35df2/packages/utils/constants/date.ts#L1
* type: https://github.com/silverlogic/baseapp-frontend/blob/c29cb34c59309b44a110cdb8baa35e6871c35df2/packages/utils/types/date.ts#L6

packages/components/modules/payments/web/InvoiceListTable/components/InvoiceItemWrapper.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return smDown ? (
```
anicioalexandre: would be nice to use if/return instead of ternary, just in case we have a third component to return (loading state, etc...)

packages/components/modules/payments/web/InvoiceListTable/index.tsx
anicioalexandre: and it would be nice to avoid ternary operatosr inside the view (UI) part as much as we can

anicioalexandre (changes requested):

anicioalexandre (approved):

PR #287, opened 2025-08-21
feature/BA-2493-available-subscriptions

lbjunq:
✅ Stripe Checkout - Subscription Plans Page
Description

    As a User, on the BaseApp Profile Page,I would like to view and manage available subscription plans on a responsive page tailored for different devices, In order to select or manage a plan that meets my needs effectively.

 
Acceptance Criteria 
Frontend

    Responsive Layouts

        Desktop (Horizontal layout):

            Display plans in a row, account for multiple rows if more than 3 plans exist.

            Ensure proper alignment and consistent card widths regardless of the number of plans.

        Mobile & Tablet (Vertical layout):

            Stack plans vertically to ensure usability and readability on smaller screens.

CoP comments

     

        Its ok to reuse Stripe Web Elements. 

        Check if its feasible to replace the BaseApp Components with the Stripe Payment Elements

            Check so that we can customize the individual component so that the end result is as similar as possible to BaseApp components in the designs: https://docs.stripe.com/elements/appearance-api#rules

        We have to think about what customizations are we going to be enabled so that components can be re used in other projects.

            Probably a good idea to talk to CoP when thinking of this.

Plan Cards

    Reusable plan card component:

        Title, pricing, description, and features dynamically populated via backend.

Call-to-Action Buttons

    Highlight the “Subscribe” button for available plans.

        Use of the button to redirect user to Checkout Page will be covered in a follow-up story 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=7170-121730&t=EqObBvkTBQ7zD0Fr-4 

Approvd 
https://app.approvd.io/projects/BA/stories/38018 

Demo: https://www.loom.com/share/b024bb2566c34682bd0c182616248275

## Summary by CodeRabbit

- New Features
  - Browse available subscriptions, toggle monthly/yearly, view pricing, features, and active status.
  - Complete Stripe-based checkout: address entry, payment method selection/addition, error handling, and success confirmation.
  - Manage payment methods: list, set default, remove, and add new via modal.
  - Subscription management: view plan details, change plan, update default payment method, and cancel with confirmation; includes Free plan view.
  - Invoices table with pagination, mobile-friendly rows, and receipt links.
  - New payment card icons (Visa/Mastercard) and improved dialog with optional hidden Cancel button.


packages/components/modules/payments/web/AvailableSubscriptions/SubscriptionCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onManageClick,
+  onSubscribeClick,
+}: SubscriptionCardProps) => {
+  const dolarPrice = Math.floor((sub.defaultPrice?.unitAmount ?? 0) / 100)
+  const centsPrice = ((sub.defaultPrice?.unitAmount ?? 0) % 100).toString().padStart(2, '0')
+
+  return (
+    <SubscriptionCardWrapper key={sub.id} smDown={smDown}>
+      {sub?.images?.length > 0 && (
+        <Image src={sub.images[0] ?? ''} alt={sub.name} width={28} height={28} />
```
mathieubouhelier: is there any risk of passing an empty string ('')?

packages/components/modules/payments/web/AvailableSubscriptions/SubscriptionCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <Button variant="contained" color="inherit" onClick={onSubscribeClick}>
+          Subscribe
+        </Button>
+      )}
+      <Box display="flex" flexDirection="column" gap={2}>
+        {sub.marketingFeatures?.map((feature) => (
+          <Typography variant="body2" color="text.secondary" key={feature.name}>
+            <CheckMarkIcon color="inherit" /> {feature.name}
+          </Typography>
+        ))}
```
mathieubouhelier: Are we sure to always get marketingFeatures?
What about `sub?.marketingFeatures?.map`

packages/components/modules/payments/web/AvailableSubscriptions/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          {yearlySubs?.length && yearlySubs?.length > 0 && (
+            <ToggleButton value="yearly">Yearly</ToggleButton>
+          )}
+        </ToggleButtonGroup>
+      </Box>
+      <Box display="flex" gap={2} width="100%" height="100%" flexWrap="wrap">
+        {selectedProducts?.map((product) => {
+          const isActive = customer?.subscriptions?.find(
+            (subscription) =>
+              subscription.status === 'active' && subscription.productsIds.includes(product.id),
```
mathieubouhelier, line 65: What bout this 
`subscription?.productsIds?.includes(product?.id ?? '') === true`
Just to avoid breaking if one undefined comes up?

packages/components/modules/payments/web/PaymentMethodsManagementComponent/index.tsx
```diff
@@ -219,10 +219,13 @@ const PaymentMethodsManagementComponent: FC<PaymentMethodsManagementComponentPro
 const PaymentMethodsManagementComponentWithElements = ({
   entityId,
   stripePublishableKey,
-}: PaymentMethodsManagementComponentWithElementsProps) => (
-  <Elements stripe={getStripePromise(stripePublishableKey)}>
-    <PaymentMethodsManagementComponent entityId={entityId} />
-  </Elements>
-)
+}: PaymentMethodsManagementComponentWithElementsProps) => {
+  console.log('stripePublishableKey', stripePublishableKey)
```
mathieubouhelier: could we remove this consolelog?

packages/components/modules/payments/web/SubscriptionManagement/FreePlanComponent.tsx
```diff
@@ -0,0 +1,59 @@
+import { Check } from '@mui/icons-material'
+import { Box, Button, List, ListItem, ListItemIcon, ListItemText, Typography } from '@mui/material'
+import { AppRouterInstance } from 'next/dist/shared/lib/app-router-context.shared-runtime'
+
+import { RowFlexContainer, SubscriptionPlanContainer } from './styled'
+
+const FreePlanComponent = ({ router }: { router: AppRouterInstance }) => {
+  const freeFeatures = [
```
mathieubouhelier, line 8: Do we have any interest in moving it to utils or constants?

mathieubouhelier (approved):

packages/components/modules/payments/web/AvailableSubscriptions/index.tsx
lbjunq: Thanks to BE setup, this should not be a possibility.

packages/components/modules/payments/web/PaymentMethodsManagementComponent/index.tsx
lbjunq: Done

packages/components/modules/payments/web/SubscriptionManagement/FreePlanComponent.tsx
lbjunq: I don 't believe this is reutilized anywhere, so for now we can leave it here

packages/components/modules/payments/web/AvailableSubscriptions/SubscriptionCard.tsx
lbjunq: No, there is not

packages/components/modules/payments/web/AvailableSubscriptions/SubscriptionCard.tsx
lbjunq: Good point, handled that better

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const [lastAddedPaymentMethodIdDuringSession, setLastAddedPaymentMethodIdDuringSession] =
+    useState<string | null>(null)
+
+  const [isAddCardModalOpen, setIsAddCardModalOpen] = useState(false)
+  const [isCancelSubscriptionModalOpen, setIsCancelSubscriptionModalOpen] = useState(false)
+
+  const queryClient = useQueryClient()
+  const invalidateCustomer = () => {
+    queryClient.invalidateQueries({ queryKey: [CUSTOMER_API_KEY.get(), entityId] })
+  }
```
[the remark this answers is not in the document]
anicioalexandre: @lbjunq coderabbit was spot on here, we should use the` CUSTOMER_API_KEY.get` to return the key, if it can change with the entityId, we adapt the  `CUSTOMER_API_KEY.get` function to receive it, that way we return already the list with the `entityId`, instead of repeating this every time

I noticed we split the stripe service and its keys into a separate file (`packages/components/modules/payments/web/services/keys.ts`)
I got the idea, but still, I think we should keep having `ApiClass` (e.g `StripeApi`) and its keys on the same file. So if we had another service file different from stripe, we won't mix the two `ApiClass`'s keys into just one `keys.ts` file

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
```diff
@@ -1,59 +1,263 @@
...
+      },
+    },
+  )
+
+  const amountDue = (subscription?.upcomingInvoice?.amountDue ?? 0) / 100
+  const nextPaymentAttempt = formatDateFromApi(subscription?.upcomingInvoice?.nextPaymentAttempt)
+  const isSubscriptionActive = subscription?.status === 'active'
+  const hasActiveSubscription =
+    customer?.subscriptions?.length &&
+    customer.subscriptions.length > 0 &&
```
anicioalexandre, line 91: Isn't this condition a bit redundant?

lbjunq: I don't belive it is, as:
* it checks for the array to be preset/loaded;
*  if it's undefined/false, as if it is, it will crash the comparison with 0;
* and only when everything is present/loaded, check for the status

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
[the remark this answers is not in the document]
anicioalexandre: @lbjunq we should be addressing that

anicioalexandre (changes requested):

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
anicioalexandre: I mean, if `customer?.subscriptions?.length` is truthy, it would mean already is not zero (thus higher than zero). So the second condition wouldn't be needed. But anyway, doesn't hurt

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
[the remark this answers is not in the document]
lbjunq: I got the key usage point, and will address it, but I disagree with the keys file point.
I think it's clearer to keep it in a separate file then to just clug the stripe file with it. I would be ok if the suggestion is to change it from 'keys.ts' to 'stripe_keys.ts', but to merge it with the other file feels not great code organization.

anicioalexandre (approved):
let's fix the CI and then we're good to merge :)

lbjunq:
CI was failing previous as well due to unsync with master of the epic. I will address it in the next story, that is the wrapup for the stripe flow

PR #288, opened 2025-08-27
User Deletion - Remove User Info from Messages

mathieubouhelier:
Acceptance Criteria 
As a user in the Messaging Feature, I want messages from deleted users to preserve their content but not show their name, avatar, or profile link, so that the conversation history remains readable without exposing data from deleted accounts.

Acceptance Criteria  

When a user deletes their account, the system should anonymize their presence in all chat messages they participated in.  

The user's name and avatar should be removed from all past messages.  

Replace the avatar with a generic "Deleted User" icon.  

Display a label such as “Deleted User” instead of their name.  

The link to the deleted user's profile should be removed from all messages.  

Messages sent by deleted users should remain visible to other participants in the chat.  

Group and direct chats should continue to display messages from deleted users with anonymized attribution.  

The message layout and order should remain unchanged after account deletion. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Falls back to “Deleted User” when a sender/profile is missing while preserving intentionally empty names.
  * Shows avatars for messages with an explicitly null profile and for first messages in grouped threads from others.
  * Ensures conversation headers display “Deleted User” and omit avatars when the other participant no longer exists.
  * Improves consistency across message lists and group views, reducing visual glitches.

### DIFF
--- a/packages/components/modules/messages/web/MessagesList/MessagesGroup/UserMessage/index.tsx
+++ b/packages/components/modules/messages/web/MessagesList/MessagesGroup/UserMessage/index.tsx
@@ -59,10 +59,10 @@ const UserMessage: FC<UserMessageProps> = ({
   }, [allMessages, allMessagesLastIndex, messageIndex])
 
   const isOwnMessage = currentProfile?.id === message?.profile?.id
-
+  const isProfileNull = message?.profile === null
   const flexAlignments = isOwnMessage ? 'flex-end' : 'flex-start'
-  const canShowAvatar = isFirstGroupedMessage && !isOwnMessage
 
+  const canShowAvatar = isProfileNull || (isFirstGroupedMessage && !isOwnMessage)
   const canShowName = canShowAvatar && isGroup
 
   if (!message) return null
@@ -92,7 +92,7 @@ const UserMessage: FC<UserMessageProps> = ({
       >
         {canShowName && (
           <Typography variant="subtitle2" color="text.primary" marginBottom={1 / 2}>
-            {message?.profile?.name}
+            {message?.profile?.name ?? 'Deleted User'}
           </Typography>
         )}
         <MessageItem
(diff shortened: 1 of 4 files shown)


lbjunq (approved):

joaosoutto (approved):

packages/components/modules/messages/web/MessagesList/MessagesGroup/UserMessage/index.tsx
```diff
@@ -92,7 +92,7 @@ const UserMessage: FC<UserMessageProps> = ({
       >
         {canShowName && (
           <Typography variant="subtitle2" color="text.primary" marginBottom={1 / 2}>
-            {message?.profile?.name}
+            {message?.profile?.name ?? 'Deleted User'}
```
anicioalexandre: wouldn't this be a bit misleading? since, in theory (looking at the profile model), a name can be null. Maybe we just check if the profile itself is null/undefined? Then we assume the user got deleted...
cc: @nossila

lokmanliton:
Deleted User's avatar is not updated with deleted user icon.
image

mathieu-tsl:
> Deleted User's avatar is not updated with deleted user icon. image

@lokmanliton , The deleted user icon isn’t included in the current design. Once it’s provided, we can add it.

packages/components/modules/messages/web/MessagesList/MessagesGroup/UserMessage/index.tsx
mathieu-tsl: @anicioalexandre Totally agree. 
Code updated with a check  `{isProfileNullOrUndefined ? 'Deleted User' : message?.profile?.name}`

anicioalexandre (approved):

nossila (approved):

PR #290, opened 2025-09-10
feature/BA-2737-profiles-multiples-members

lbjunq:
✅ Multiple Profiles - Members Suggestion Dropdown

Description

    As a user, on the BaseApp Profile,I would like to I want to add new members to my organization, In order to have thse members help in the organizations activites.

 

Original Story:  
Acceptance Criteria 
Context
This story is about creating the Add members button and the search functionality to find users in the platform to invite them as members. 
The suggestions dropdown should be generic so that in other projects the queries could be changed. So that we can suggest other stuff that is not users.
We are going to use the user query just for the sake of using an example.

    Sidebar with AA

Business Rules - Members Suggestion Dropdown
On this story we will work on the search functionality to add members, but not the actual logic of the invitation.

    Given a user wants to add a new member to the organization, when they click the Add Member button, a modal should appear where they can input the name or the user name. 

    The input box should suggest a list of users, based on what the user has written in it.

        The filtering criteria for this search will be name of the user.

    We will show only 4 options max and then use an infinite scrolling behavior to expand the initial results.

    Given the I have found the corresponding user, when I click on the user option, it will be added to the list of users to invite after confirmation.

    Given the user has clicked on Invite Button, the invited users should now appear in the Members list page with the Label invited. 

Demo: https://www.loom.com/share/d991c36180c847d2b34dd82cf116a22d


packages/components/modules/profiles/web/ProfileMembers/MembersList/components/AddMembersDialog.tsx
```diff
@@ -0,0 +1,275 @@
...
+    const isNewEmail = 'isNewEmail' in option
+    return (
+      <UserListItemContainer
+        key={isNewEmail ? option.email : (option as User).id}
+        onClick={() => {
+          handleItemSelection(option)
+        }}
+      >
+        {isNewEmail ? (
+          <Box
```
mathieu-tsl: what do you think moving this to styled?

mathieu-tsl (approved):

packages/components/modules/profiles/web/ProfileMembers/components/MemberPersonalInfo.tsx
anicioalexandre, line 1: could we follow the file hierarchy and structure?
https://bitbucket.org/silverlogic/baseapp-frontend-template/src/master/CODE-GUIDELINES.md#component-hierarchy
https://bitbucket.org/silverlogic/baseapp-frontend-template/src/master/CODE-GUIDELINES.md#component-hierarchy

in that case, there is no need for a components folder since MemberPersonalInfo should be implicitly a component of ProfileMembers

packages/components/modules/profiles/web/ProfileMembers/MembersList/components/AddMembersDialog.tsx
anicioalexandre, line 1: could we follow the file hierarchy and structure?
https://bitbucket.org/silverlogic/baseapp-frontend-template/src/master/CODE-GUIDELINES.md#component-hierarchy
https://bitbucket.org/silverlogic/baseapp-frontend-template/src/master/CODE-GUIDELINES.md#component-hierarchy

in that case, there is no need for a components folder since AddMembersDialog should be implicitly a component of MembersList

the same applies to UserCard and VirtuosoListBox.
lets make sure each one gets it own types file as well

anicioalexandre (changes requested):

anicioalexandre (approved):

anicioalexandre (changes requested):
hold

PR #294, opened 2025-10-02
Feature/ba 2713 stripe flow wrapup

lbjunq:
❌ Subscription Page Component Integration
Description fields
Description

Acceptance Criteria 

    Implement components to the subscription page so that everything is integrated with existing components in baseapp. 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/42883 

Demo:
https://www.loom.com/share/0f76db0716b6466ebae525250630c2c6


lbjunq:
Originally I was also planning on doing the merge with master here, but since this is already quite big, I'll do it in a follow up.

mathieubouhelier (approved):

pablogadhi (approved):

packages/components/modules/payments/web/PaymentDropDown/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      setSelectedPaymentMethodId(selectedId)
+    }
+  }
+
+  useEffect(() => {
+    if (isCreatingSetupIntent || isErrorCreatingSetupIntent) return
+    if (setupIntent) {
+      // @ts-ignore
+      elements?.update({ clientSecret: setupIntent.clientSecret })
```
anicioalexandre: Let's do this on `onSuccess` of `useSetupIntent` instead of doing that on a `useEffect`

packages/components/modules/payments/web/PaymentDropDown/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    createSetupIntent()
+  }
+
+  const handleCloseModal = () => {
+    setIsAddCardModalOpen(false)
+  }
+
+  const handleSelectPaymentMethod = (e: SelectChangeEvent<string>) => {
+    const selectedId = e.target.value as string
+    if (selectedId === 'add-new') {
```
anicioalexandre: Let's put this on a constant and reuse

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
```diff
@@ -37,12 +38,12 @@ import {
...
 import { getChipLabelAndColorByStatus } from './utils'
 
-const SubscriptionManagement: FC<SubscriptionManagementProps> = ({ entityId, router }) => {
+const SubscriptionManagement: FC<SubscriptionManagementProps> = ({ entityId }) => {
   const [lastAddedPaymentMethodIdDuringSession, setLastAddedPaymentMethodIdDuringSession] =
     useState<string | null>(null)
-
   const [isAddCardModalOpen, setIsAddCardModalOpen] = useState(false)
   const [isCancelSubscriptionModalOpen, setIsCancelSubscriptionModalOpen] = useState(false)
+  const [subscriptionId, setSubscriptionId] = useState<string | null>(null)
```
anicioalexandre, line 46: why do we need one more state? cant we use useGetSubscription data?

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
```diff
@@ -123,8 +131,20 @@ const SubscriptionManagement: FC<SubscriptionManagementProps> = ({ entityId, rou
...
   }, [lastAddedPaymentMethodIdDuringSession])
 
-  if (!hasActiveSubscription && !isLoading) {
-    return <FreePlanComponent router={router} />
+  useEffect(() => {
+    if (customer?.subscriptions?.[0]?.id) {
+      setSubscriptionId(customer.subscriptions[0].id)
+      refetchSubscription()
+    }
+  }, [customer])
```
anicioalexandre, line 142: do we really need this effect? if we use `useGetSubscription`, then all we need to do is make sure that it receives the `customer.subscriptions[0].id`, then whenever this value changes it should also update the query state, since we use the sub id on the query key

packages/components/modules/payments/web/hooks/useStripeHook.tsx
anicioalexandre, line 1: we should make so the `queryKey` matches the `queryFn` related key, the way we currently have is a bit confusing, like he currently do:
```
const useGetCustomer = (entityId?: string) =>
    useQuery({
      queryKey: [CUSTOMER_API_KEY.get(entityId?.toString() ?? 'me')], <--- instead of using a query key abstraction directly we are creating a new list with the `CUSTOMER_API_KEY.get`
      queryFn: () => StripeApi.getCustomer(entityId),
    })`
 
```

Ideally, we would have something like this:
```
// on services/stripe.ts
class StripeApi {
  static getCustomer = (entityId?: string): Promise<Customer> =>
    axios.get(`${baseUrl}/stripe/customers/${entityId ?? 'me'}`)
  ...
  }
  
export const CUSTOMER_API_KEY = {
  default: ['customer'],
  getCustomer: (entityId?: string) => [...CUSTOMER_API_KEY.default, 'getCustomer', entityId],
}

// usage
useQuery({
    queryKey: CUSTOMER_API_KEY.getCustomer(entityId),
    queryFn: () => StripeApi.getCustomer(entityId),
})`
```

This way, the queryKey pretty much reflects the fetch function

Can we refactor those keys at `packages/components/modules/payments/web/services/keys.ts` so they follow the same idea? Like, if you see we have SUBSCRIPTION_API_KEY, CREATION_SUBSCRIPTION_API_KEY, UPDATE_SUBSCRIPTION_API_KEY... since they all belong to `StripeApi` class, it should probably be only one object, then we make sure the keys reflect the `StripeApi` class methods instead

Also would be nice to move the keys to `packages/components/modules/payments/web/services/stripe.ts`, so we keep keys closer to the service itself (and avoid having multiple service keys altogether on the same keys.ts file)

packages/components/modules/payments/web/hooks/useStripeHook.tsx
```diff
@@ -137,14 +137,13 @@ const useStripeHook = () => {
...
+  const useCreateSubscription = () =>
     useMutation({
       mutationFn: (options: CreateSubscriptionOptions) => StripeApi.createSubscription(options),
       onSuccess: () => {
-        console.log('Subscription created successfully:')
-      },
-      onError: (error) => {
-        sendToast(`Failed to create subscription: ${error.message}`, { type: 'error' })
+        queryClient.invalidateQueries({
+          queryKey: [CUSTOMER_API_KEY.get(), SUBSCRIPTION_API_KEY.get()],
```
anicioalexandre: we should have a queryKey for `StripeApi.getSubscription` method, then we can reuse it on the mutations that need it like `useCreateSubscription`, `useUpdateSubscription`....

anicioalexandre (changes requested):

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
lbjunq: This was necessary because this information is being pulled off the customer, but even though the customer is updated, the subscription fetch is not being updated on page change.
It was a fix for the following situation:
Subscribed to a product.
Went to profile settings, all good, updated correctly.
Change to another product.
When you return to profile settings again, even though customer was refetched, subscription was not, even on a refetch call it would still show the previous one.

So I transformed it into a state to properly track the change and refetch correctly.
I'm open to alternatives, but couldn't quite do another one, specially considering budget has already been reached for this.

packages/components/modules/payments/web/SubscriptionManagement/index.tsx
lbjunq: Answered in the comment above

packages/components/modules/payments/web/services/stripe.ts
```diff
@@ -10,11 +12,67 @@ import {
...
+// export const CANCEL_SUBSCRIPTION_API_KEY = {
+//   default: ['useCancelSubscription'] as QueryKey,
+//   get: (...params: string[]) => [...CANCEL_SUBSCRIPTION_API_KEY.default, ...params] as QueryKey,
+// }
+
+export const SETUP_INTENT_API_KEY = {
+  default: ['useSetupIntent'] as QueryKey,
+  get: (...params: string[]) => [...SETUP_INTENT_API_KEY.default, ...params] as QueryKey,
+}
+
```
anicioalexandre, line 48: the idea would be having something like this:


```
export class StripeApi {
  ...
}

export const STRIPE_API_KEY = {
  default: ['stripe'],
  getCustomer: (entityId?: string) =>
    [...STRIPE_API_KEY.base,'getCustomer',entityId] as const,
  // createCustomer: (...)
  // createSetupIntent: (...)
  // listPaymentMethods: (...)
  // updatePaymentMethod: (...)
  // deletePaymentMethod: (...)
  // listInvoices: (...)
  // listProducts: ()
  // getProduct: (...)
  // createSubscription: (...)
  // getSubscription: (...)
  // updateSubscription: (...)
  // cancelSubscription: (...)
}
}
```

The goal is to have query keys that directly reflect the service methods, so the cache key and the fetch function are always aligned and easy to reason about

anicioalexandre (changes requested):

packages/components/modules/payments/web/services/stripe.ts
lbjunq: Are you sure this would be the best method? I get it, but it would make the invalidate default a little too powerful/broad, as with it slightly split as we have it right now, the default is more useful and localized.

anicioalexandre: > but it would make the invalidate default a little too powerful/broad

that's why we add the method string into the query key (e.g. 'getCustomer') so if you want to invalidate just the method, no matter what the params, you can

It makes sense for the query keys to be centralized  per API Class method, this shouldn't be an issue, really. Let's follow that for now, since all the other parts follow it too, then we can potentially re-evaluate it

anicioalexandre (changes requested):

anicioalexandre:
@lbjunq can we fix build/lint?

anicioalexandre (approved):

PR #296, opened 2025-10-06
feature/BA-26-26-Single chat components

gabrieleandro:
Single chat creation components


packages/components/modules/messages/native/SingleChatCreate/CreateRoomList/CreateRoomListItem/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <Pressable key={`chat-room-item-${node?.id}`} onPress={handleRoomCreation}>
+      <View style={styles.cardContainer}>
+        <View>
+          <AvatarWithPlaceholder imgSource={node?.image?.url} />
+        </View>
+        <View>
+          <Text variant="subtitle2">{node?.name}</Text>
+          <Text variant="caption">
+            {node?.urlPath?.path && `@${node?.urlPath.path?.replace('/', '')}`}
```
anicioalexandre: didnt get why we the backend would return the / at the end, was this always like that? If so at this point, I think we should probably create a util so we can reuse it on the template, since it is becoming a bit repetitive. Maybe we could place that util at the [shared module](https://github.com/silverlogic/baseapp-frontend/tree/master/packages/components/modules/__shared__/common)

packages/components/modules/messages/native/SingleChatCreate/CreateRoomList/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  useEffect(() => {
+    layoutTriggeredRef.current = false
+    startTransition(() => {
+      refetch({ q: searchParam })
+    })
+  }, [refetch, searchParam])
+
+  return (
+    <View style={styles.flatListWrapper}>
+      <FlatList
```
anicioalexandre: lets  reuse the `InfiniteScrollerView` component like we did on the [notifications feature](https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/notifications/native/NotificationsList/index.tsx#L78)

packages/components/modules/messages/native/SingleChatCreate/CreateRoomList/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    }
+    return null
+  }
+
+  useEffect(() => {
+    layoutTriggeredRef.current = false
+    startTransition(() => {
+      refetch({ q: searchParam })
+    })
+  }, [refetch, searchParam])
```
anicioalexandre, line 45: most of the times it's cleaner (and less buggy) to have a more event-based approach than relying on these effects. Like in this case, if we could call `refetch` on the `SearchInput`'s `onChangeText` instead of watching `searchParam` change and triggering `refetch`. Maybe a TODO so we can think about it when we have time?
https://react.dev/learn/you-might-not-need-an-effect

anicioalexandre: I dont think I got the layoutTriggeredRef thingy, what is it solving?

anicioalexandre (approved):

PR #301, opened 2025-11-12
feature/BA-2725 profile summary drawer

rodean1:
Acceptance Criteria 
As a user in a one-on-one chat, I want to open a profile summary drawer when I click the other participant’s name or avatar, so that I can quickly view their details without leaving the chat.

Acceptance Criteria  

Clicking the participant’s name or avatar should trigger the opening of a profile summary drawer component from the side of the screen.  

Given I open the hamburger menu on the top right, I should see a contact details option. When I click it, it should open the profile summary drawer as well.

The profile summary drawer must retrieve and display:  

Display name  

Avatar  

username

Bio  

The drawer component should be able to be closed

The drawer should be displayed while keeping the conversation visible and usable.

Data for the profile summary must be fetched from the backend using the correct profile ID associated with the current conversation.

The profile summary drawer should include a clearly visible “Go to Profile” action that navigates to the participant’s full profile page.  

It should open a new tab

The navigation to the profile page should open the correct identity for the participant

Implementation should be responsive for Desktop, tablet and mobile view

Guardrails

No animation is needed for the drawer 

Design Link: https://www.figma.com/design/XRD6wSl1m8Kz6XUcAy5CLp/BaseApp---WEB?node-id=8109-145277&t=U2v8Soqr9Bh1UMZw-0 

Approvd 
https://app.approvd.io//projects/BA/stories/43040 2

## Summary by CodeRabbit

* **New Features**
  * "Contact Details" menu item opens a Profile Summary panel with avatar, name, username, biography, and actions ("Go to Profile", "Add Contact to Group").
  * Profile Summary view added with back navigation to return to chat list.

* **UX**
  * Header can trigger the Profile Summary; profile data is preloaded for faster display.

* **Style**
  * New profile icon added for visual consistency.


packages/components/modules/messages/common/context/useChatRoom/constants.ts
```diff
@@ -4,4 +4,5 @@ export const LEFT_PANEL_CONTENT = {
   createGroupChat: 2,
   editGroupChat: 3,
   groupDetails: 4,
+  profileSummary: 5,
 } as const
```
anicioalexandre, line 8: lets move this closer to the `useChatRoom`/`ChatRoomProvider` context, since we moved this logic there

packages/components/modules/messages/common/context/ChatRoomProvider/types.ts
```diff
@@ -9,6 +17,14 @@ type ChatRoomFunctions = {
     replace?: boolean | undefined,
   ) => void
   resetChatRoom: () => void
+  setLeftPanelContent: (content: number) => void
```
anicioalexandre: lets use:
`export type LeftPanelContentValues = ValueOf<typeof LEFT_PANEL_CONTENT>` (move this type over to here)
instead of number

packages/components/modules/messages/web/ChatRoom/ChatRoomHeader/index.tsx
```diff
@@ -37,14 +38,28 @@ const ChatRoomHeader: FC<ChatRoomHeaderProps> = ({
...
-  const { resetChatRoom } = useChatRoom()
+  const { resetChatRoom, setSingleChatProfileDetails, setLeftPanelContent } = useChatRoom()
 
   const { isGroup } = roomHeader
-  const { title, avatar } = useNameAndAvatar(roomHeader)
+  const { title, avatar, path, pk, biography } = useChatroomDetails(roomHeader)
+
+  useEffect(() => {
+    if (!isGroup) {
+      setSingleChatProfileDetails({
```
anicioalexandre: I think that the best approach here would be the `ProfileSummary` itself calling its own Fragment to get the data it needs and then register this fragment under `GroupDetailsQuery` defined at `ChatRoomsComponent`.
By doing that we dont need to store anything in this state provider and we keep in sync with the relay philosophy in which we specify the data requirements of a component by using fragments

packages/components/modules/messages/web/ProfileSummary/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      />
+    </>
+  )
+}
+
+const SuspendedProfileSummary: FC<ProfileSummaryProps> = ({ onBackButtonClicked, ...props }) => (
+  <Suspense
+    fallback={
+      <>
+        <Header onBackButtonClicked={onBackButtonClicked} />
```
anicioalexandre, line 32: nice!

packages/design-system/components/web/icons/ProfileNoCircleIcon/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC } from 'react'
+
+import { SvgIcon, SvgIconProps } from '@mui/material'
+
+const ProfileNoCircleIcon: FC<SvgIconProps> = ({ sx, ...props }) => (
+  <SvgIcon sx={{ fontSize: 24, color: 'grey.800', ...sx }} {...props}>
+    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path
+        d="M12 11.1998C13.1 11.1998 14.2 10.7998 15 9.9998C16.7 8.2998 16.7 5.5998 15 3.9998C14.2 3.1998 13.1 2.7998 12 2.7998C10.9 2.7998 9.79995 3.1998 8.99995 3.9998C7.29995 5.6998 7.29995 8.3998 8.99995 9.9998C9.79995 10.7998 10.9 11.1998 12 11.1998ZM10.1 5.0998C10.6 4.5998 11.3 4.2998 12 4.2998C12.7 4.2998 13.4 4.5998 13.9 5.0998C15 6.1998 15 7.8998 13.9 8.9998C12.9 9.9998 11.1 9.9998 9.99995 8.9998C8.99995 7.8998 8.99995 6.0998 10.1 5.0998Z"
+        fill="#637381"
```
anicioalexandre: we usually change the fill to currentColor to allow it to be reused with different colors (only works with icons with only one color)
https://github.com/silverlogic/baseapp-frontend/blob/master/packages/design-system/components/web/icons/README.md

packages/components/modules/messages/web/ChatRoomsComponent/index.tsx
```diff
@@ -31,15 +32,14 @@ const ChatRoomsComponent: FC<ChatRoomsComponentProps> = ({
   GroupChatEditComponentProps = {},
   SingleChatCreateComponent = DefaultSingleChatCreate,
   SingleChatCreateComponentProps = {},
+  ProfileSummaryComponent = DefaultProfileSummary,
 }) => {
   const isUpToMd = useResponsive('up', 'md')
-  const [leftPanelContent, setLeftPanelContent] = useState<LeftPanelContentValues>(
```
anicioalexandre, line 36: nice move! 🎉

packages/components/modules/messages/web/ChatRoom/ChatRoomHeader/ChatRoomOptions/index.tsx
```diff
@@ -11,11 +11,17 @@ const ChatRoomOptions: FC<ChatRoomOptionsProps> = ({
   onArchiveClicked,
   onDetailsClicked,
   onLeaveClicked,
+  onContactDetailsClicked,
 }) => (
   <MenuList>
     <MenuItem onClick={onArchiveClicked} disabled={isArchiveMutationInFlight}>
       <Typography variant="body2">{isArchived ? 'Unarchive Chat' : 'Archive Chat'}</Typography>
     </MenuItem>
+    {!isGroup && (
```
Ronan-Fernandes: We are checking `isGroup` twice here, the first checks if it is not a group, and the second is a ternary that returns null if it is not a group.

So maybe instead of checking this twice, we could just paste the not group part in the second part of the ternary below.

packages/components/modules/messages/web/ProfileSummary/Body/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const Body: FC<BodyProps> = ({ avatarSize = 144, chatRoomRef }) => {
+  const { currentProfile } = useCurrentProfile()
+  const profileSummary = useFragment<ProfileSummaryFragment$key>(
+    ProfileSummaryFragment,
+    chatRoomRef,
+  )
+  const getSingleChatDetails = () => {
+    const details = profileSummary?.participants?.edges?.map((edge) => {
+      if (edge?.node?.profile?.id !== currentProfile?.id) {
```
Ronan-Fernandes, line 36: what happens if `edge?.node?.profile?.id` returns undefined? 
If i'm not mistaken i think if that is undefined and current id exists this will always evaluate to true 

and then you endup with all the object values with undefined as well.

I'm not sure if this is a behavior we want.
if not maybe we could save profile in a variable and check if profile and profile id exists before comparing with current id.

If we do that i think we wont have this issue where the object values are all undefined.

Please check this

anicioalexandre (approved):
good job!

PR #316, opened 2025-12-13
feature/BA-2703-update-group-chat-participants

rodean1:
Related BE PR: https://github.com/silverlogic/baseapp-backend/pull/357

Acceptance Criteria 
Current Behavior: https://www.loom.com/share/add12dcb7bfb4531ad28e52d00ec9aed

Adding New Members

Given an admin adds new participants to the group chat, the new participants should be added to the list without the need to refresh.

The list should follow the same sorting rules it has now

Fix Edit Mode Searchbar

Given I am editing a group chat, when input text in the search bar, then the participant list should filter to match the search bar query. 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/42979 

## Summary by CodeRabbit

* **New Features**
  * Added member removal flow with a confirmation dialog and admin checks; member removal can now be initiated from member lists.
  * Improved member list pagination for chat rooms (variable-driven participants loading).

* **UI/UX Improvements**
  * Updated removal action label to "Remove from chat".
  * Confirm button loading/disabled states simplified for add/remove member dialogs.

* **Chores**
  * Release bumped to 1.4.17; changelog entry added.


Ronan-Fernandes (approved):

packages/components/modules/messages/common/graphql/queries/GroupDetailsQuery.ts
```diff
@@ -1,13 +1,13 @@
...
-  query GroupDetailsQuery($roomId: ID!) {
+  query GroupDetailsQuery($roomId: ID!, $q: String) {
     chatRoom(id: $roomId) {
       id
       participantsCount
       ...GroupTitleFragment
-      ...MembersListFragment
-      ...ProfileSummaryFragment
+      ...MembersListFragment @arguments(q: $q)
+      ...ProfileSummaryFragment @arguments(q: $q)
```
nossila: it seems weird to have `MembersListFragment` and `ProfileSummaryFragment` listed here, and they all have essentially the same fields, I think we should fully remove `ProfileSummaryFragment` from the code base, and also look into removing `RoomTitleFragment`, we should move this logic to the backend and treat group and 1-1 room's name and avatar the same way on the frontend.

@anicioalexandre what do you think?

anicioalexandre: @nossila  super agreed, I noticed we have some repeated fragments that we could merge too

rodean1: can this be added to the follow up story that @nossila is writing for the technical debt? I already closed the BE pull request for this story

packages/components/modules/messages/web/GroupChatEdit/index.tsx
```diff
@@ -118,21 +136,36 @@ const GroupChatEdit: FC<GroupChatEditProps & { profileId: string }> = ({
...
+    if (profile?.id) {
+      setRemovedMemberIds((prev) => [...prev, profile.id])
+    }
+  }
+
   const handleAddMemberSuccess = () => {
     setOpen(false)
     startTransition(() => {
-      refetch?.({})
+      refetch(
```
nossila: I think we should look into leverage relay's mutation response and/or optimistic update instead of filter by removed users and force a refetch... it is not required to refetch and with optimistic update we can make it feel very fast and we don't need to manually manipulate the data, also the code will be easier to maintain and will be a good example for other cases.

@anicioalexandre what do you think?

nossila (changes requested):

packages/components/modules/messages/web/GroupChatEdit/index.tsx
anicioalexandre: Agreed, we have some examples already using optimistic updates instead. Could you check it @rf2tsl ? Should be on the messages web implementation

packages/components/modules/messages/common/graphql/queries/GroupDetailsQuery.ts
anicioalexandre: do we already have the follow-up @nossila?

anicioalexandre (changes requested):

packages/components/modules/messages/common/graphql/queries/GroupDetailsQuery.ts
rodean1: @anicioalexandre yes we have one already, you can add to it and increase the points since I haven't started work on it yet

https://silverlogic.atlassian.net/browse/BA-2949

anicioalexandre (approved):

nossila (approved):

PR #317, opened 2025-12-15
feature/BA-2940-rn-edit-comment

lbjunq:
✅ RN - Edit a Comment
Key details
Description

Acceptance Criteria 

    Given that a user has created a comment, when they press and hold the comment, then a drawer should slide up from the bottom displaying the comment options menu.

        Implement a backdrop when the drawer comes up.

    When the user selects the Edit Comment option from the menu, then an input box should open, pre-filled with the current comment content.

    The input box should label informing the user that they are editing an existing comment.

    Given that the user is editing a comment, when they click the "X" on the label, then the edit should be canceled, and the comment should remain unchanged.

    Given that the user has finished editing the comment, when they press the send button, then the updated comment should be saved.

    When a comment has been edited and saved, then an "Edited" label should appear next to it, indicating that the original content has been modified. 

Design Link: https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7245-44177&t=pkyMShfd1d7IXixb-0 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/39784 

Demo: https://www.loom.com/share/76dd3129d79a4a0fa2b35973b38ad3cd


packages/components/modules/comments/native/CommentItem/index.tsx
```diff
@@ -26,43 +37,163 @@ const CommentItem: FC<CommentItemProps> = ({
...
+              gap: 8,
+              padding: 8,
+            }}
+          >
+            <PinIcon width={20} height={20} color={theme.colors.object.high} />
+            <Text variant="body2" color="high">
+              Pin Comment
+            </Text>
+          </Pressable>
+          <Pressable
```
nossila: @luciobj can we only show those buttons if the user has the correct permission?

The check/query is already available for the comment instance https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/comments/common/graphql/queries/CommentItem.ts#L37

And you can see an usage example here https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/comments/web/CommentItem/useCommentOptions/index.tsx#L43

packages/components/modules/comments/native/Comments/index.tsx
```diff
@@ -20,6 +28,24 @@ const WithComments: FC<CommentsProps> = ({
   CommentCreate = DefaultCommentCreate,
   CommentCreateProps = {},
 }) => {
+  const [isEditMode, setIsEditMode] = useState(false)
+
+  const form = useForm<SocialUpsertForm>({
```
nossila, line 55: looks strange to hold a form instance in `Comments` component, even if you never going to edit a comment 🤔  I think it is better to leave the form in `CommentCreate` like it was before and control/manipulate that as necessary, or maybe even duplicate that component if required

packages/components/modules/comments/native/CommentContainer/index.tsx
```diff
@@ -50,6 +43,26 @@ const CommentCreate = forwardRef<NativeTextInput, CommentCreateProps>(
...
+            }
+            form.reset()
+            if (ref && 'current' in ref) ref.current?.blur()
+          },
+        })
+        isEdit?.onEditCancel()
+        return
+      }
 
       commitMutation({
```
nossila: do we need to execute `commitMutation` even if it is for edit? If we are going to have both comment create mutation and update mutation inside this component then we should at least rename it `CommentCreate` is not a good name

nossila (changes requested):

packages/components/modules/comments/native/CommentContainer/index.tsx
lbjunq: We don't need to execute `commitMutation` if it's for edit, which is why we have the early return. I will change the component name as suggested though

nossila (approved):

packages/components/modules/comments/native/CommentItem/index.tsx
```diff
@@ -26,43 +37,169 @@ const CommentItem: FC<CommentItemProps> = ({
...
-      </View>
-    </View>
+      </Pressable>
+      <BottomDrawer
+        bottomDrawerRef={bottomDrawerRef}
+        handleSheetChanges={handleSheetChanges}
+        snapPoints={['30%']}
+      >
+        <View
+          style={{
```
anicioalexandre: lets move this to a `styles.ts` file and create  a stylesheet for that, we could reuse a lot of styles here, like on the Pressable component
example: https://github.com/silverlogic/baseapp-frontend/blob/master/packages/components/modules/messages/native/MessageCreate/styles.ts

packages/components/modules/__shared__/native/SocialInput/types.ts
```diff
@@ -19,4 +22,5 @@ export interface SocialInputProps {
   onTextHeightChange?: (height: number) => void
   shouldUseBottomSheetSafeComponents?: boolean
   submit: VoidFunction
+  isEdit?: SocialTextInputIsEditProps
```
anicioalexandre: `isEdit` is a tricky name here, we often think of it as a boolean instead of an object. Can we rename it?

anicioalexandre (changes requested):
it would have been nice if we could have separated the rebase with the PR, it gets a lot harder to review :/

packages/components/modules/comments/native/Comments/index.tsx
```diff
@@ -1,44 +1,266 @@
...
+  const handleMenuAction = useCallback(
+    (_action: string) => {
+      bottomDrawerRef.current?.close()
+      if (_action === 'edit') {
+        handleEdit(selectedComment as CommentItem_comment$data)
+      }
+      setSelectedComment(undefined)
+    },
+    [selectedComment, handleEdit],
+  )
```
anicioalexandre, line 174: doesn't this over complicates the `Comments` component? Since it's supposed to just assemble all the parts.
Couldn't we pass the mutations to the components that execute them? like the edit being at the CommentUpdate (could be similar to the [web version](https://github.com/silverlogic/baseapp-frontend/tree/master/packages/components/modules/comments/web/CommentUpdate))
then the submit part could be at CommentCreate...

anicioalexandre (changes requested):

packages/components/modules/comments/native/Comments/index.tsx
lbjunq: I don't think we could do similar to web, due to design differences. Mobile utilizes the same Input for all changes, be it edit or reply, and web creates the input within the comment you are trying to do the action.
By centralizing the logic here, we also don't duplicate the bottom drawer menu.

nossila: @anicioalexandre I had the same though as you the first time I reviewed it, but I approved because we need bigger changes to achieve that in a nice way, so I think we are better at having it working and shipped and have another story to refactor it in a better way

also this is targeting an epic branch, so we can refactor before merging to master as well

anicioalexandre: @lbjunq the web also uses the same input for the actions, but it has this wrapper abstraction like CommentCreate where we put the methods in

packages/components/modules/__shared__/common/constants.ts
```diff
@@ -4,10 +4,12 @@ import { SocialUpsertForm } from './types'
 
 export const DEFAULT_SOCIAL_UPSERT_FORM_VALUES: SocialUpsertForm = {
   body: '',
+  id: '',
```
anicioalexandre, line 7: why would the id be inside the form? didnt get this part

anicioalexandre (approved):
I'll aprove it but hopping that we can reestructure it before we merge the epic

packages/components/modules/__shared__/common/constants.ts
lbjunq: This is for the mutation

PR #318, opened 2025-12-22
Feature/ba 2964 rn reply comment and reply list

lbjunq:
✅ RN - Replying to Comments
Key details
Description

Acceptance Criteria 
User Story
As a user I want to reply to a comment, so that I can engage in discussions and respond directly to specific comments

Business rules

    Given a comment has already been created, a reply button should appear next to the reactions button.

    Given the user clicks on the reply button, the mobile keyboard should slide up in order for the user to type in the content.

        The input box will slide up with the keyboard, and it should remain visible.

        A label should be displayed communicating that you are replying to the owner of the comment.

    Given I click the X button in the label, then we should cancel the reply creation.

    Given a user has finished creating a reply, when they press the "send button", then the reply should be created and appear at the top of the reply list.

    Each reply should display the user’s avatar, name, and timestamp

    Each reply should also display the reaction button.

    Given the reply has been published successfully, then expand the reply list up to 5 most recent additional replies (6 in total = 5 most recent + the 1 new reply). 

Design Link: https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7220-109867&t=jbYKzqIHwPMDjdiS-0 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/39782 

Demo: https://www.loom.com/share/5a4c3a6c021b4866a72bc137d68006fe


lbjunq:
Since this PR is based on [another one](https://github.com/silverlogic/baseapp-frontend/pull/317), preferably wait for that one to be merged before reviewing this.

[Edit]: This is now ready to be reviewed

anicioalexandre:
@lbjunq please resolve conflicts and rebase with the epic so we can review

masfiqurtsl (approved):

dpTsl:
Screenshot from 2026-01-20 11-47-50

Then you created a new reply:

Screenshot from 2026-01-20 11-48-07


"This is a new comment..." seems to be duplicate. I think it wasn't suppose to be displayed at the top of comments list. Is that correct?

dpTsl:
I'm not sure if this should be tackled in this story or not, but once the replies are expanded the "show more replies" button disappear. https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7233-167206&t=GKtGGXejNownsp7E-0

And the "Hide replies" button appear at the bottom of the replies list
https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7233-167256&t=GKtGGXejNownsp7E-0

dpTsl (approved):

lbjunq:
> Screenshot from 2026-01-20 11-47-50
> 
> Then you created a new reply:
> Screenshot from 2026-01-20 11-48-07
> 
> "This is a new comment..." seems to be duplicate. I think it wasn't suppose to be displayed at the top of comments list. Is that correct?

This is likely due to the current implementation, which is not accurate. There shouldn't be a page for comments, and instead they should be 'attributes' of pages or posts. So this bug can be something that will not carry to the final implementation. In any case, I've taken note of it, and will make sure it's not in the epic merge.

> I'm not sure if this should be tackled in this story or not, but once the replies are expanded the "show more replies" button disappear. https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7233-167206&t=GKtGGXejNownsp7E-0
> 
> And the "Hide replies" button appear at the bottom of the replies list https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7233-167256&t=GKtGGXejNownsp7E-0

This is gonna be handled by a follow up story: https://silverlogic.atlassian.net/browse/BA-2382

packages/components/modules/comments/native/CommentItem/index.tsx
```diff
@@ -9,73 +9,160 @@ import { useRefetchableFragment } from 'react-relay'
...
+              // eslint-disable-next-line no-console
+              console.error('Error loading replies:', error)
+            } else {
+              setIsRepliesExpanded(true)
+            }
+          },
+        },
+      )
+    }
+  }, [commentIdToExpand, comment.id, isRepliesExpanded, hasReplies, refetch])
```
anicioalexandre, line 94: why do we need these side effects? I think we should keep it event-based (expand when the button is clicked) like the web implementation...
what's the intention here? to auto expand when a comment is created?

packages/components/modules/comments/native/Comments/index.tsx
```diff
@@ -83,13 +99,24 @@ const WithComments: FC<CommentsProps> = ({
...
     if (isCreateMutationInFlight || isUpdateMutationInFlight) return
 
     nextClientMutationId += 1
     const clientMutationId = nextClientMutationId.toString()
 
-    const connectionID = ConnectionHandler.getConnectionID(target.id, 'CommentsList_comments')
+    const connectionID =
+      replyVariables?.isReplyMode && id
+        ? ConnectionHandler.getConnectionID(id, 'CommentsList_comments')
+        : ConnectionHandler.getConnectionID(target.id, 'CommentsList_comments')
```
anicioalexandre, line 118: we shouldn't need this if we split components into `CommentCreate` and `CommentUpdate`
This is just a small example... I feel we should think about to rearchitecture this components following what we have on the web. The way we have kept driving us to different implementations, I'm afraid the effort to change this once the epic is done will be quite considerable...
@nossila thoughts?

packages/components/modules/comments/native/Comments/index.tsx
```diff
@@ -132,7 +160,13 @@ const WithComments: FC<CommentsProps> = ({
         }
         const mutationErrors = response?.commentCreate?.errors
         setFormRelayErrors(form, mutationErrors)
-
+        if (replyVariables?.isReplyMode) {
+          replyVariables?.onReplyCancel()
+          if (!mutationErrors?.length && id) {
+            setCommentIdToExpand(id)
+            setTimeout(() => setCommentIdToExpand(null), 100)
```
anicioalexandre, line 167: why do we need this?

anicioalexandre (changes requested):

packages/components/modules/comments/native/CommentItem/index.tsx
lbjunq: Exacly. One of the AC's was to expand the list whenever a reply is made. This was how I achieved that

packages/components/modules/comments/native/Comments/index.tsx
lbjunq: Same thing as before, to achieve expansion upon creating a comment, then clearing the field should we reply under another comment

packages/components/modules/comments/native/CommentItem/index.tsx
nossila: On web we expand the replies before you can reply, so no side effect needed.

To be really honest, I'm not a fan of that behavior, it not so explicit for the user how to reply there...

@lbjunq what if you just reuse `showReplies` function? You can make the `fetchPolicy` an argument of that function, that way you can call if with the proper policy on both cases... you can get rid of the side effect by calling `showReplies` whenever the person clicks on the button?

nossila: @lbjunq or change the default `fetchPolicy` to 'store-and-network` this will get you both worlds, the caching for faster loads and refresh from network

packages/components/modules/comments/native/CommentItem/index.tsx
```diff
@@ -9,73 +9,160 @@ import { useRefetchableFragment } from 'react-relay'
...
+
+  const renderCommentsReplies = () => {
+    const CommentsListComponent = (
+      <RepliesList
+        target={comment as CommentsList_comments$key}
```
nossila: Use `commentRef` here instead of `comment` and you wont need to change the type... if feel like you need to to any `as` for a `$key` type, there is something wrong.

nossila (changes requested):

packages/components/modules/comments/native/Comments/index.tsx
nossila: @anicioalexandre I agree with you that we should split into different components, I don't like that we have mutations on the top Comments component, but I also feel like this rearchitecture is not related to this PR, in my opinion we need to rethink other stuff in this epic, even on the web there is some rooms for improvement, like for exemple on web it is not clear you can reply to a comment, there is no "reply" button, but anyway, in my opinion we should merge and finish this epic with some changes here and there to get the feature done, then we rethink the architecture and maybe the designs to improve the architecture as well... after it is done we can rearchitecture it to better reuse logic on both web and mobile, it is totally possible, I just don't like a lot the idea to hold stuff because it is not perfect... those are my current thoughts.

anicioalexandre: I also don't wanna hold this, I'm just afraid that rearchitecting at the end of the epic will be a big effort and we might not have the time it deserves to do it

anicioalexandre (approved):
approving, but we should review a few things on this implementation later on

nossila (approved):

PR #321, opened 2026-01-05
feature/BA-2930-rn-messages-group-create-new-group-chat

davidbermudez-tsl:
Acceptance Criteria 
Create new Group Chat

Given I click on the "Create New Chat" button, when I select the option to create a new group chat, then I should be navigated to the group creation screen.

Given I am on the group creation screen and enter a name for the group, when I save the group, then the group name should be saved and displayed at the top of the chat.

Given I am on the group creation screen and upload a profile picture for the group, when the upload is successful, then the picture should appear as the group’s icon.

Given I have entered a group name , when I confirm and save the group, then the group should be created and appear in my active chat list.

The Group name is mandatory.

if no profile picture was uploaded then display a placeholder image.

Given I am creating a group chat, when I access the participant selection section, then I should be able to see a list of all my contacts.

Given I am creating a group chat and the contact list is displayed, when I use the search bar, then I should be able to filter the contact list to find specific contacts.

Reuse the Search bar component we implemented in the Chat Lists.

Use an infinite scroller

We will have 2 different lists:

Members: are your contacts that have will be added to the group.

Contacts: your contacts that could be added to the group.

Given I press the "add" button of a contact, then it should move to the members section of the list.

Given I have added participants and named the group, when I confirm the creation, then the group chat should be created and appear in my active chat list.

The Save button should only be enabled after the Group Name has a value and at least 1 participant besides myself has been added.

We always want to create a new group chat regardless if we already have an existing room for the same participants. 

Design Link: https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7770-220223&t=qZVBT6H3IAyD7hH8-0 

Notes 

Small improvement story - for adding the names of the individual in the group who sent the message if not possible to cover within this story 

https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=7829-36894&t=0yeOQaFlNi1pvAzM-0 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/41472 

## Summary by CodeRabbit

* **New Features**
  * Create group chats with participant selection, member preview/removal, and Create Group flow.
  * Group details screen: set group name, upload/remove group avatar with upload progress and feedback.

* **UI Updates**
  * New Groups tab in messaging and New Chat flows route to group creation when active.
  * AppBar close control supports accessible label and disabled state.
  * Added picture and check-mark icons.


packages/components/modules/messages/common/graphql/mutations/CreateGroupChat.ts
```diff
@@ -0,0 +1,147 @@
...
+            const response = await FileSystem.uploadAsync(
+                process.env.EXPO_PUBLIC_RELAY_ENDPOINT as string,
+                fileUri,
+                {
+                    httpMethod: 'POST',
+                    uploadType: FileSystem.FileSystemUploadType.MULTIPART,
+                    fieldName,
+                    parameters: {
+                        // TODO: see if we can reuse the query from the mutation
+                        query: `mutation ChatRoomUpdateMutation($input: ChatRoomUpdateInput!) {
```
nossila: it is possible to reuse a compiled query, check on `baseapp-frontend-template/apps/web/components/pages/app-router/page.tsx` for an example

davidbermudez-tsl: I don't think so because this is a mutation

ac-tsl (approved):

packages/components/modules/messages/native/ChatRoomPage/MessagesList/MessagesGroup/UserMessage/index.tsx
```diff
@@ -32,8 +36,14 @@ const UserMessage: FC<UserMessageProps> = ({
 
     return isPreviousMessageFromOtherParticipant || roomHasOnlyOneMessage
   }, [allMessages, allMessagesLastIndex, messageIndex])
-
-  const isOwnMessage = currentProfile?.id === message?.profile?.id
+  const { me } = useLazyLoadQuery<ProfileCurrentUserQueryType>(
```
nossila: it is not a good idea to use `useLazyLoadQuery` on a component that might be rendered multiple times in the same page like this one

You should be able to access  `use.profile.id` from `useJWTUser`, but you need to double check: if `use.profile.id` is an integer then we will need to change the backend to return the profile's global id as well

But I also think we should keep the logic as it was before, check only against the current profile, because of the nature of the multiple profiles feature, what was the reason to change this logic?

packages/components/modules/messages/native/SingleChatCreate/CreateRoomList/index.tsx
```diff
@@ -1,26 +1,44 @@
...
+    },
+  )
+
   const [isPending, startTransition] = useTransition()
   const { data, refetch, loadNext, isLoadingNext, hasNext } = useAllProfilesList(targetRef)
 
-  const profiles = data?.allProfiles?.edges ?? []
+  const profiles =
+    data?.allProfiles?.edges?.filter((profile) =>
+      isGroup ? profile?.node?.id !== me?.profile?.id : true,
```
nossila: I think we should be using `currentProfile` here instead of `me.profile`

nossila (changes requested):

nossila (approved):

packages/design-system/components/native/appbars/AppBar/index.tsx
```diff
@@ -34,14 +44,41 @@ const AppBar: FC<AppBarProps> = ({
           {title}
         </Text>
       )}
-      <View style={styles.buttonContainer}>
```
anicioalexandre, line 37: @davidbermudez-tsl I'm pretty sure that there was a reason to render this view regardless of whether the `onClose` is passed. If I remember well, it makes the title components always centered between routes that have and don't have the `onClose`. If we don't render the `buttonContainer` view when `onClose` not used, the title will move a tiny bit between routes.

I think that, for the `next >` use case we need, we could've just overwritten the entire component using the `closeComponent` prop at the template level (`apps/mobile/app/(protected-routes)/create-group/_layout.tsx`), and that would solve it, I think. I know the naming is a bit odd, though. Maybe we could rename it from `closeComponent` to `rightComponent` or something like it?

I don't think it makes sense to have `onNext` and `onClose` since they overlap and do the same thing essentially, we could aim into something more generic, what do you think?

anicioalexandre (changes requested):

packages/design-system/components/native/appbars/AppBar/index.tsx
davidbermudez-tsl: It's all gonna be as "close" instead of "next" now, and also left the buttonContainer

packages/design-system/components/native/appbars/AppBar/index.tsx
```diff
@@ -34,12 +39,32 @@ const AppBar: FC<AppBarProps> = ({
...
+      <View style={[styles.buttonContainer, { minWidth: 24, width: 'auto' }]}>
```
anicioalexandre: Could we make these styles into the styles.ts sheet class? were trying to keep the component as clean as we can

packages/design-system/components/native/appbars/AppBar/index.tsx
```diff
@@ -34,12 +39,32 @@ const AppBar: FC<AppBarProps> = ({
...
-              <CloseIcon />
-            </IconButton>
+            <Pressable
+              disabled={closeDisabled}
+              accessibilityRole="button"
+              accessibilityLabel={closeLabel || 'Close'}
+              accessibilityHint="Close the current screen"
+              onPress={onClose}
+            >
+              <View style={{ flexDirection: 'row', alignItems: 'center' }}>
```
anicioalexandre: same here

anicioalexandre (approved):

PR #326, opened 2026-01-11
feature/BA-2949-profile-settings-update

rodean1:
Acceptance Criteria 
Currently the settings are under /user/settings and all children pages are built using a state to identify the current tab, if you are at members management page and reloads, you get sent back to the edit profile tab.

To fix that, lets change the implementation:

Convert  the tabs are actual routes

Then lets change the useLazyLoadQuery implementation to make use of usePreloadedQuery instead.

Lets change the profile settings router to be /profile/:id/settings instead of /user/settings

Query for `canChange: hasPerm(perm: "change")` and if the user loads the profile settings and it returns `false` then we should show a permission required page instead. 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/44399 

## Summary by CodeRabbit

* **New Features**
  * Profiles now include a primary key (pk) across auth, UI, and profile-related screens to improve profile identification and propagation.
  * Profile settings and switching now react to pk changes, ensuring current profile updates reliably across the app.

### DIFF
--- a/packages/authentication/modules/profile/useCurrentProfile/__tests__/__mock__/profiles.ts
+++ b/packages/authentication/modules/profile/useCurrentProfile/__tests__/__mock__/profiles.ts
@@ -6,6 +6,5 @@ export const mockUserProfileFactory = (id: string) => {
     name: faker.person.fullName(),
     image: faker.image.avatar(),
     urlPath: faker.internet.url(),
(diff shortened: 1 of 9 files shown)


rodean1:
no longer needed, recent backend changes refactor id to uuid and exposing pk is probably not a good idea anyway

PR #328, opened 2026-01-18
feature/BA-2949 profile settings tab permissions check

rodean1:
Acceptance Criteria 
Currently the settings are under /user/settings and all children pages are built using a state to identify the current tab, if you are at members management page and reloads, you get sent back to the edit profile tab.

To fix that, lets change the implementation:

Convert  the tabs are actual routes

Then lets change the useLazyLoadQuery implementation to make use of usePreloadedQuery instead.

Lets change the profile settings router to be /profile/:id/settings instead of /user/settings

Query for `canChange: hasPerm(perm: "change")` and if the user loads the profile settings and it returns `false` then we should show a permission required page instead. 

Approvd 
https://app.approvd.io/silverlogic/BA/stories/44399 

## Summary by CodeRabbit

* **New Features**
  * Permission-based access control added to profile pages — components now hide when you lack required permissions.
  * Users see a clear "Permission Required" screen with a "Go Back" button when access is denied.
  * Profile loading now respects permission checks to prevent unauthorized component rendering.


nossila (approved):

packages/components/modules/profiles/common/graphql/queries/Profile.ts
```diff
@@ -3,6 +3,7 @@ import { graphql } from 'react-relay'
 export const ProfileByIdQuery = graphql`
   query ProfileByIdQuery($id: ID!) {
     profile(id: $id) {
+      canChange: hasPerm(perm: "change")
```
anicioalexandre: maybe add canDelete too?
based on the settings route we could maybe have:
- credentials          -> canChange/no perms?
- delete-account   -> canDelete
- edit-profile          -> canChange
- members             -> no perms?
- notifications        -> no perms?
- theme                  -> no perms?

packages/components/modules/profiles/web/CheckPermissionWrapper/types.ts
```diff
@@ -0,0 +1,6 @@
+import { ReactNode } from 'react'
+
+export interface ProfilePermissionWrapperProps {
+  target: any
```
anicioalexandre: lets type this using the fragment type

packages/components/modules/profiles/common/graphql/queries/Profile.ts
```diff
@@ -3,6 +3,7 @@ import { graphql } from 'react-relay'
 export const ProfileByIdQuery = graphql`
   query ProfileByIdQuery($id: ID!) {
     profile(id: $id) {
+      ...CheckMountPermissionWrapper @arguments(perm: "delete")
```
anicioalexandre: Are we hardcoding "delete" here? Shouldn't it be an argument too? So other pages can change the permission, something like this?
```
export const ProfileByIdQuery = graphql`
  query ProfileByIdQuery($id: ID!, $perm: String! = "change") {
    profile(id: $id) {
      ...CheckMountPermissionWrapper @arguments(perm: $perm)
      ...ProfileComponentFragment
      ...FollowToggleUpdatableFragment
    }
  }
`
```

anicioalexandre (changes requested):

anicioalexandre (approved):

nossila (approved):

PR #334, opened 2026-01-27
BA-2961 third party sign in google

ficast:
This should be merged after BA-3008


packages/authentication/modules/access/allauth/utils.ts
```diff
@@ -68,3 +68,19 @@ export function normalizeAllAuthError(error: unknown) {
...
+      sessionToken: undefined,
+    }
+  }
+
+  return {
+    accessToken: response.meta.accessToken,
+    refreshToken: response.meta.refreshToken,
+    sessionToken: response.meta.sessionToken,
+  }
+}
```
Hercilio1: It doesn't seem that you are using this function anywhere. I also believe it's not needed. Let's remove it?

Hercilio1 (changes requested):

packages/authentication/modules/access/allauth/useAllAuthGoogleLogin/index.ts
```diff
@@ -0,0 +1,60 @@
...
+ *   }
+ * })
+ *
+ * // Note: You should use GoogleLogin button component or useGoogleOneTapLogin
+ * // instead of calling loginWithGoogle directly
+ * <GoogleLogin onSuccess={loginWithGoogle} />
+ * ```
+ */
+const useAllAuthGoogleLogin = ({ backendLoginOptions = {} }: UseAllAuthGoogleLoginOptions = {}) => {
+  // MUTATION: Send Google id_token to backend
```
priscilladeroode: The code does not seem too complex, so the comments are not needed.

packages/authentication/modules/access/index.ts
```diff
@@ -22,6 +22,12 @@ export { default as useAllAuthSignUp } from './allauth/useAllAuthSignUp'
 export { default as useAllAuthRecoverPassword } from './allauth/useAllAuthRecoverPassword'
 export { default as useAllAuthResetPassword } from './allauth/useAllAuthResetPassword'
 export { default as useAllAuthLogout } from './allauth/useAllAuthLogout'
+export { default as useAllAuthSession } from './allauth/useAllAuthSession'
+export { default as useAllAuthGoogleLogin } from './allauth/useAllAuthGoogleLogin'
+export type * from './allauth/useAllAuthGoogleLogin/types'
+export { useAllAuthGoogleLoginHandler } from './allauth/useAllAuthGoogleLoginHandler'
+export type { UseAllAuthGoogleLoginHandlerOptions } from './allauth/useAllAuthGoogleLoginHandler'
+export { extractTokensFromAllAuthResponse, isAllAuthPasswordChangeRedirect } from './allauth/utils'
```
priscilladeroode: Seems like this extractTokensFromAllAuthResponse is not being used, can you remove?

priscilladeroode (changes requested):

packages/authentication/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@hookform/resolvers": "catalog:",
+    "@react-oauth/google": "^0.13.4",
```
anicioalexandre: if we are keeping this dependency here and on the template, we might want to move it to the pnpm shared catalog

Hercilio1 (approved):

priscilladeroode (approved):

PR #354, opened 2026-03-30
Improve Comments Customizations

Ronan-Fernandes:
- `__package_name__` package update - `v __package_version__`
  - **changelog_info**
  - **changelog_info**

## Summary by CodeRabbit

* **New Features**
  * Added BaseComments for advanced comments integration
  * Added opt-in comment sharing control

* **Documentation**
  * Clarified provider usage to avoid double-wrapping; when to use BaseComments

* **Refactor**
  * Made comment and input UI components more customizable (overridable form/container, per-reply prop passthrough, profile-ID toggle)

* **Style**
  * Switched comment/input backgrounds to use theme default background for better theming consistency

* **Tests**
  * Simplified comments delete test flow (removed prior negative assertion)


packages/components/modules/comments/web/CommentItem/index.tsx
```diff
@@ -196,6 +202,13 @@ const CommentItem: FC<CommentItemProps> = ({
           subscriptionsEnabled={subscriptionsEnabled}
           onReplyClick={onReplyClick}
           CommentItem={CommentItem}
+          // Make sure the replies are also customizable with the same props as the parent comment item
+          CommentItemProps={{
+            enableDelete,
+            enableShare,
+            useProfileId,
+            CommentContainer,
+          }}
```
anicioalexandre: Good catch. We should do it for all the others, right? maybe something like:
https://github.com/silverlogic/baseapp-frontend/pull/355/changes/9d80a9fea05fc2012b7c42f1441db230ef3f2ec9

packages/components/modules/comments/web/CommentItem/index.tsx
```diff
@@ -31,11 +31,14 @@ const CommentItem: FC<CommentItemProps> = ({
...
   ActionOverlayProps,
   enableDelete = false,
+  enableShare,
   Timestamp = DefaultTimestamp,
   CommentUpdate = DefaultCommentUpdate,
   CommentReplyButton = DefaultCommentReplyButton,
   CommentPinnedBadge = DefaultCommentPinnedBadge,
   CommentReactionButton = DefaultCommentReactionButton,
+  CommentContainer = DefaultCommentContainer,
+  useProfileId = false,
```
anicioalexandre: shouldnt this default to true so it's backward compatible?

packages/components/modules/comments/web/BaseComments/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  CommentsList = DefaultCommentsList,
+  CommentsListProps,
+  CommentCreate = DefaultCommentCreate,
+  CommentCreateProps,
+  onCommentCreateFocus,
+  commentCreateRef: commentCreateRefProp,
+}) => {
+  const { resetCommentReply } = useCommentReply()
+  const internalCommentCreateRef = useRef<HTMLInputElement>(null)
+  const commentCreateRef = commentCreateRefProp ?? internalCommentCreateRef
```
anicioalexandre, line 42: I wonder if we do something like:
```
const BaseComments: FC<BaseCommentsProps> = ({
  target: targetRef,
  subscriptionsEnabled = true,
  CommentsList = DefaultCommentsList,
  CommentsListProps,
  CommentCreate = DefaultCommentCreate,
  CommentCreateProps,
  onCommentCreateFocus,
  commentCreateRef=null,
}) => {
  const { resetCommentReply } = useCommentReply()
  const commentCreateRef = useRef<HTMLInputElement>(commentCreateRef)
```
would work?

packages/components/modules/comments/web/BaseComments/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { FC, useEffect, useRef } from 'react'
+
+import { useFragment } from 'react-relay'
+
+import { CommentsFragmentQuery, useCommentReply } from '../../common'
+import DefaultCommentCreate from '../CommentCreate'
+import DefaultCommentsList from '../CommentsList'
+import { BaseCommentsProps } from './types'
+
+const BaseComments: FC<BaseCommentsProps> = ({
```
anicioalexandre, line 30: That's a great solution! I like it
Could we just add a little JSDoc on this component? something like:
```
  /**
   * BaseComments renders the comments component without the `CommentReplyProvider`.
   *
   * Use this instead of `Comments` when you need access to `useCommentReply`
   * in a parent component (e.g., to programmatically control comment replies
   * from your project). In that case, wrap your parent component with
   * `withCommentReplyProvider` to ensure the context is available.
   *
   * If you don't need `useCommentReply` outside of this component, use
   * `Comments` directly — it already includes the provider.
   *
   * @example
   * // In your project, when you need useCommentReply in a parent:
   * const MyCommentsPage: FC = () => {
   *   const { setCommentReply } = useCommentReply()
   *   return <BaseComments target={targetRef} />
   * }
   * export default withCommentReplyProvider(MyCommentsPage)
   */
```

packages/components/modules/comments/web/CommentItem/index.tsx
Ronan-Fernandes: if this is default true we will start using the ProfileId instead of the profileUrlPath. currently BA uses profileUrlPath and fallback to ProfileId when profileUrlPath is not available

packages/components/modules/comments/web/BaseComments/index.tsx
Ronan-Fernandes: I tried but did not managed to make it work

packages/components/modules/comments/web/CommentItem/index.tsx
Ronan-Fernandes: just did that, thanks for the comment

packages/components/modules/comments/web/Comments/types.ts
```diff
@@ -1,15 +1,3 @@
...
-export interface CommentsProps {
-  target: CommentsFragment$key
-  subscriptionsEnabled?: boolean
-  CommentsList?: FC<CommentsListProps>
-  CommentsListProps?: Partial<CommentsListProps>
-  CommentCreate?: FC<CommentCreateProps>
-  CommentCreateProps?: Partial<CommentCreateProps>
-  onCommentCreateFocus?: (ref: HTMLInputElement) => void
-}
+export type CommentsProps = BaseCommentsProps & {}
```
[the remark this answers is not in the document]
anicioalexandre: @Ronan-Fernandes lets remove this one

anicioalexandre (approved):

anicioalexandre:
@Ronan-Fernandes could you run the component test packages locally to see if passes? it should be a flaky test
Friendly reminder to version this PR before publishing

PR #363, opened 2026-04-21
Add header center component

Ronan-Fernandes:
- `__package_name__` package update - `v __package_version__`
  - **changelog_info**
  - **changelog_info**

## Summary by CodeRabbit

* **New Features**
  * Headers now support an optional centered content area. Render custom content (e.g., search, branding, or controls) in the header center alongside the account menu.
  * All navigation layouts (centered, horizontal, mini, vertical) forward this center region and preserve account menu alignment so the account area does not shrink when center content is present.

### DIFF
--- a/packages/components/modules/navigations/web/Header/index.tsx
+++ b/packages/components/modules/navigations/web/Header/index.tsx
@@ -6,11 +6,12 @@ import { MenuIcon } from '@baseapp-frontend/design-system/components/web/icons'
 import { Logo } from '@baseapp-frontend/design-system/components/web/logos'
 import { useUISettings } from '@baseapp-frontend/design-system/hooks/web'
 
+import Box from '@mui/material/Box'
 import IconButton from '@mui/material/IconButton'
 import Toolbar from '@mui/material/Toolbar'
 
 import DefaultAccountMenu from './AccountMenu'
-import { CustomAppBar } from './styled'
+import { CustomAppBar, HeaderCenterContainer as DefaultHeaderCenterContainer } from './styled'
 import { HeaderProps } from './types'
 
 const Header: FC<HeaderProps> = ({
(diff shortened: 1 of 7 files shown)


packages/components/modules/navigations/web/NavigationLayout/index.tsx
```diff
@@ -60,6 +61,7 @@ const NavigationLayout: FC<NavigationLayoutProps> = ({
               ...AccountMenuProps,
               additionalComponent: notificationsComponent ?? AccountMenuProps?.additionalComponent,
             }}
+            HeaderCenterComponent={HeaderCenterComponent}
```
anicioalexandre, line 64: I think on nav centered this would be a bit weird no? since it has the items in the center of the header (same place you would put the search input on hive)

packages/components/modules/navigations/web/Header/index.tsx
```diff
@@ -60,7 +62,17 @@ const Header: FC<HeaderProps> = ({
...
+          <>
+            <Box sx={{ flex: 1, display: 'flex', justifyContent: 'center', minWidth: 0 }}>
+              {HeaderCenterComponent}
+            </Box>
+            <Box sx={{ flexShrink: 0 }}>
+              <AccountMenu {...AccountMenuProps}>{children}</AccountMenu>
+            </Box>
+          </>
+        )}
+        {!HeaderCenterComponent && <AccountMenu {...AccountMenuProps}>{children}</AccountMenu>}
```
anicioalexandre: let's extract this into an if/return function, so it's easier to visualize each case
Let's use styled on the first box since it has more than 3 sx, and it might be useful to be able to override it anyway

anicioalexandre (approved):

pt-tsl (approved):

PR #377, opened 2026-05-26
fix: include groups in active/unread chat tabs, overlay tab loading spinner and last message preview

nossila:
- Drop `defaultValue: false` on `isGroup` in `RoomsListFragment` so unset means "no filter" (returns both group and non-group rooms).
- In `AllChatRoomsList`, send `isGroup: true` only on the Groups tab and `null` on Active / Unread / Archived, so groups also appear in those tabs.
- Update `useRoomListSubscription` connection matcher to treat a `null`, `isGroup` on a stored connection as "match any room", so newly arrived group rooms slot into Active / Unread / Archived connections.
- Render the tab loading `CircularProgress` overlaid on top of the tab label (LoadingButton-style) instead of pushing it to the side, keeping the tab width stable while refetching.
- Render chat room last message as clear text preview

## Summary by CodeRabbit

* **New Features**
  * Group rooms now correctly appear when filtering Active, Unread, and Archived tabs.
  * Message previews simplified for cleaner, capped text shown in room list.

* **Bug Fixes**
  * Tab loading spinner now overlays the tab label without shifting width.

* **Documentation**
  * Components package updated to v1.6.0 and changelog entry added.

[Review Change Stack](https://app.coderabbit.ai/change-stack/silverlogic/baseapp-frontend/pull/377?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/components/modules/messages/web/AllChatRoomsList/ChatRoomItem/utils.ts
```diff
@@ -8,6 +8,55 @@ import {
...
+
+  for (let i = 0; i < MAX_HTML_STRIP_ITERATIONS && current !== previous; i += 1) {
+    previous = current
+    current = current.replace(/<[^>]{0,500}>/g, '')
+  }
+
+  return current.replace(/[<>]/g, '')
+}
+
+const stripMarkdownSafely = (line: string) =>
```
anicioalexandre, line 16: why do we want to remove the md @nossila ? I thought we had added the markdown preview everywhere on this messages feature. Or is there a reason not to show parsed markdown?

packages/components/modules/messages/web/AllChatRoomsList/ChatRoomItem/utils.ts
```diff
@@ -8,6 +8,55 @@ import {
...
+    .replace(/^\s*\d+\.\s+/, '')
+    .trim()
+
+export const getLastMessagePreview = (content?: string | null) => {
+  if (!content) return ''
+
+  const safeContent = content.slice(0, MAX_PREVIEW_INPUT_LENGTH)
+
+  return (
+    stripHtmlTagsSafely(
```
anicioalexandre: Is there a chance to have HTML in these messages? maybe we should use a purify library instead? like dompurify

anicioalexandre (approved):

packages/components/modules/messages/web/AllChatRoomsList/ChatRoomItem/utils.ts
nossila: @anicioalexandre it breaks preview in a number of ways, and this is the same behavior of other messaging apps so I think it is widely expected that on the list preview it only shows first line and clear text, no formatting

packages/components/modules/messages/web/AllChatRoomsList/ChatRoomItem/utils.ts
nossila: @anicioalexandre we already handle markdown, so I don't see a reason to also handle html, besides this is the preview in which case we don't want any formatting, it will only show a snippet of the message in clear text.

nossila: @anicioalexandre about dompurify, to use that we need to first render markdown -> html -> dompurify, to be honest I would love to move this to the backend instead, both the markdown -> html converter and also cache a clear text version of the latest message on save time, but for now lets keep it on the frontend and monitor performance.

PR #380, opened 2026-06-04
feature: add unarchive to groups

vitorguima:
- Adds archived groups to the "archived" tab in the chat page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed room vs. group filtering so switching tabs shows the correct list.

* **Improvements**
  * Automatically fetches latest data when returning to the app or changing tabs.
  * After a refetch, the list reliably scrolls back to the top so users see updated results.

### DIFF
--- a/packages/components/modules/messages/native/RoomsList/RoomsListComponent/index.tsx
+++ b/packages/components/modules/messages/native/RoomsList/RoomsListComponent/index.tsx
@@ -39,6 +39,16 @@ const RoomsListComponent: FC<RoomsListProps> = ({ targetRef, searchParam, select
     return undefined
   }
 
+  const shouldFilterByGroup = () => {
+    if (selectedTab === CHAT_TAB_VALUES.groups) {
+      return true
+    }
+    if (selectedTab === CHAT_TAB_VALUES.active) {
+      return false
+    }
+    return null
+  }
+
   useFocusEffect(
     useCallback(() => {
       layoutTriggeredRef.current = false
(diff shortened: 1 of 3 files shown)


packages/components/modules/messages/native/RoomsList/RoomsListComponent/index.tsx
```diff
@@ -39,6 +39,16 @@ const RoomsListComponent: FC<RoomsListProps> = ({ targetRef, searchParam, select
     return undefined
   }
 
+  const shouldFilterByGroup = () => {
+    if (selectedTab === CHAT_TAB_VALUES.groups) {
+      return true
+    }
+    if (selectedTab === CHAT_TAB_VALUES.active) {
+      return false
```
nossila: @vitorguima I think we want to return groups in the active tab as well, no? I mean, active are all currently chat rooms, groups + individuals.

I made this change in the web but didn't apply it to native: https://github.com/silverlogic/baseapp-frontend/pull/377/changes#diff-65021c4c10465a2a3d9e04a88f02277fca2db732a51a300fb2bdeefba299fbafL64

packages/components/modules/messages/native/RoomsList/RoomsListComponent/index.tsx
```diff
@@ -48,16 +58,17 @@ const RoomsListComponent: FC<RoomsListProps> = ({ targetRef, searchParam, select
             q: searchParam,
             unreadMessages: selectedTab === CHAT_TAB_VALUES.unread,
             archived: selectedTab === CHAT_TAB_VALUES.archived,
-            isGroup: selectedTab === CHAT_TAB_VALUES.groups,
+            isGroup: shouldFilterByGroup(),
           },
-          { fetchPolicy: 'store-and-network' },
+          { fetchPolicy: 'network-only' },
```
nossila: I believe you changed this to network to make the list not show outdated info, but I think there is a new thing about 'store-and-network' is that it shows the list right away, even though it might be outdated, users will see something anyway, this is whatsapp behavior as well... just a thought

nossila (approved):

anicioalexandre (approved):

PR #381, opened 2026-06-05
Feature/ba 3286 rn delete a comment

davidbermudez-tsl:
As a user who owns a comment or has permission to moderate,
I want to be able to delete a comment in the mobile app,
So that I can manage and clean up content I no longer want to be visible, just like I can on the web.

Description:
This feature enables users to delete their own comments (or comments they have permission to moderate) directly from the mobile app. The delete action should be accessible through an overlay drawer.We already have this functionality on web.

Scenario: Display delete option for a comment in the app

Given the user is viewing a comment in the mobile app, and  holds press over it, the overlay drawer displays the message "Delete Comment" in red.

Scenario: Show delete confirmation dialog before deleting comment

Given the user taps the delete action for a comment, when the system triggers the deletion flow, then a confirmation dialog appears with the message “Are you sure you want to delete this comment? This action cannot be undone.”

And the dialog includes two buttons: “Cancel” and “Delete”

And the “Delete” button is styled as a destructive action in red

Scenario: Cancel deletion from confirmation dialog

Given the confirmation dialog is visible, when the user taps “Cancel” or closes the dialog, then the system dismisses the dialog and no changes are made

Scenario: Confirm deletion from confirmation dialog

Given the confirmation dialog is visible, when the user taps “Delete”, then the system deletes the comment and removes it from the thread

And the comment is no longer visible in the comment list

Scenario: Ensure parity with web behavior

Given the delete feature works on web, when the same user attempts to delete a comment in the app, then the app provides the same functionality and confirmation flow as the web version

https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=9612-33758&t=PQHECmyjd8eFKcVb-1 

https://www.figma.com/design/z5ZdVlKEMmGIX4GVnjwfxA/BaseApp---NATIVE?node-id=9612-33049&t=PQHECmyjd8eFKcVb-1 

Approvd
https://app.approvd.io/silverlogic/BA/stories/44378

### DIFF
--- a/packages/components/modules/__shared__/common/ReactionButton/index.tsx
+++ b/packages/components/modules/__shared__/common/ReactionButton/index.tsx
@@ -1,5 +1,3 @@
-import { FC, MouseEvent } from 'react'
-
 import { useCurrentProfile } from '@baseapp-frontend/authentication'
 
 import { graphql, useFragment, useMutation } from 'react-relay'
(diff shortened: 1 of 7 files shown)


anicioalexandre (approved):
Approving the merge to the epic, but we still need to do a better review on the comments epic and fix a few things

masfiqurr (approved):

PR #392, opened 2026-06-18
feat: React Native comments — all logic shared with web via a common hook layer

nossila:
Adds the full comments experience to React Native and, in the process, restructures the comments module so **all behavior lives in `comments/common` and the platform folders are thin UI layers**. History is linear (rebased onto master, 22 commits), and every commit is green.

### Native comments (`comments/native`)
- Create, reply, edit, and delete comments from a single morphing bottom composer (one always-mounted `SocialInputDrawer`; the gorhom sheet never remounts when switching create → reply → edit).
- Long-press action sheet (share / pin / edit / delete) gated on `hasPerm` fields, with delete confirmation; pin/unpin re-sorts the list pinned-first.
- Nested threads with a depth cap (`DEFAULT_MAX_THREAD_DEPTH = 5`), infinite scrolling, show/hide replies, and auto-expand of the target thread after submitting a reply.
- Inline form validation errors on the composer (via the design-system `SocialTextInput`); transport errors toast through the shared mutation hooks.
- Tappable author avatar/name that navigate to the profile (only when the profile has a registered `urlPath` — the id fallback is a web-only route).
- Flat pages render their body as the thread's scrolling list header (`pages/native/PageComponent`).

### Shared logic layer (`comments/common`)
- `useCommentCreateForm` / `useCommentUpdateForm` — form + submit: connection-id derivation, `clientMutationId`, `setFormRelayErrors`, reset-on-success, optional mentions, optional external `form` (see composer note below).
- `useCommentItem` — per-comment state machine: replies expansion, consume-once auto-expand signal, reply targeting, deletion, profile URL. Subscribes to the reply store per-item via a `useShallow` selector, so store writes don't re-render every visible comment.
- `useCommentActions` — headless share/pin/edit/delete descriptors (labels, permission gating, pin mutation with an `onPinToggled` re-sort option); web maps them to `ActionsOverlay`, native to the action sheet.
- `useCommentList` — now returns the filtered `comments` nodes and an identity-stable `refetchWithOrder`.
- `CommentReplyProvider` store is platform-neutral (generic `commentItemRef`), gains `editingComment` (mutually exclusive with reply mode) and `commentIdToExpand`, and `useCommentReply` accepts a selector.
- Utils: `getCommentsConnectionId`, `getNextClientMutationId`, `toCommentEditTarget` (jest-covered).

### Web
`CommentCreate`, `CommentUpdate`, `CommentItem`, `useCommentOptions`, and the lists now consume the shared hooks — **props, exports, and behavior unchanged**.

### Design system
Native `SocialTextInput` (edit/reply banners, tool row, controller-driven inline field errors: error border + warning icon/caption) and five new native icons.

## Review pass

An adversarial self-review ran against the refactor diff (10 finder angles → 12 verifiers → gap sweep). All confirmed findings were fixed in-branch:
- Action sheet is always mounted (closed): gorhom's `BottomSheetModal` only opens via `present()` on a mounted ref — the first long-press in a thread used to be a silent no-op (latent bug carried over from the pre-refactor menu).
- Per-item reply-store selectors (the O(N) re-render wave on every store write is gone).
- Profile navigation gated on `urlPath` (the `/profile/<id>` fallback dead-ended on native).
- Native edit exits optimistically on submit (pre-refactor parity), and the reply button is hidden on deleted-user comments instead of silently no-oping.
- Composer drives ONE react-hook-form instance through both shared hooks: RHF's `Controller` captures `control.register` in a mount-time ref, so swapping forms made typed characters revert instantly while editing.

Intentional behavior changes vs. the pre-refactor epic: replying to a nested (depth ≥ 2) comment now auto-expands its sub-thread (previously the reply landed invisibly in an unfetched connection).

## Versioning (changesets)

- `@baseapp-frontend/design-system` **2.1.0** — minor: native SocialTextInput + icons.
- `@baseapp-frontend/components` **3.0.0** — minor changeset; the major comes from the design-system peer bump, matching this repo's release pattern.
- `@baseapp-frontend/wagtail` **3.0.0** — peer dependency bump only.

Vs. the released `components@2.0.x` everything here is additive; within the epic, native `CommentsListProps`/`CommentItemProps` dropped `onReply`/`onLongPress`/`commentIdToExpand`/`onEdit`/`target` (flows moved into the reply store and `CommentActionsProvider`), and native `CommentItem` must render inside `Comments`/`BaseComments`.

## Verification

- Relay compile with zero artifact churn at every step (no GraphQL document changes in the refactor).
- `tsc --noEmit` clean on `components` and `design-system`; eslint clean (only pre-existing warnings); jest 21/21.
- Manual native pass: composer morphing, reply auto-expand, long-press menu gating, pin re-sort, delete dialog error handling, depth cap, infinite scroll, comments-disabled fallback, profile navigation, and typing while editing.


masfiqurr:
FE looks solid for view/scroll/pin/API CRUD, Minor improvement need on Share no-op, the Relay Expected the server edge to be non-null toast, and edit exitOnSubmit swallowing errors before merge maybe.

nossila:
> FE looks solid for view/scroll/pin/API CRUD, Minor improvement need on Share no-op, the Relay Expected the server edge to be non-null toast, and edit exitOnSubmit swallowing errors before merge maybe.

@masfiqurr that "edge to be non-null" happen across the app in different places, and it is silenced/muted when running on production. I created a bug to address that in another PR: https://app.approvd.io/silverlogic/BA/stories/47217

masfiqurr (approved):

packages/components/modules/__shared__/native/Timestamp/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+const Timestamp: FC<TimestampProps> = ({ date }) => {
+  const dateTime = DateTime.fromISO(date)
+  const now = DateTime.now()
+  const diffDays = now.startOf('day').diff(dateTime.startOf('day'), 'days').days
+
+  let text
+  if (diffDays === 0) text = 'Today'
+  else if (diffDays === 1) text = 'Yesterday'
+  else if (diffDays <= 7) text = `${diffDays} days ago`
+  // TODO (Design): Can we reuse formatDateWithDiffNow or formatRelativeTeam?
```
anicioalexandre, line 19: can we check with the design team @nossila ? lower priority thou

packages/components/modules/comments/common/graphql/subscriptions/CommentsSubscription.tsx
```diff
@@ -1,6 +1,8 @@
 import { useMemo } from 'react'
 
-import { ConnectionHandler, graphql, useSubscription } from 'react-relay'
+import { graphql, useSubscription } from 'react-relay'
+
+import { getCommentsConnectionId } from '../../utils'
```
anicioalexandre, line 5: nice!

packages/components/modules/comments/native/CommentDeleteDialog/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+          onDeleted?.()
+          onClose()
+        }
+      },
+    })
+  }
+
+  return (
+    <Portal>
+      <Dialog style={styles.dialog} visible={visible} onDismiss={onClose}>
```
anicioalexandre: can we extract this Portal + Dialog combo into the design system native package? I think it would be nice to have as a default dialog component

packages/components/modules/comments/native/CommentShowRepliesButton/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const styles = createStyles(theme)
+
+  return (
+    <View style={styles.showRepliesButtonContainer}>
+      <Pressable onPress={onShowReplies} style={styles.showRepliesButton} disabled={isLoading}>
+        <Text style={styles.showRepliesButtonText}>
+          {body} {showTotalRepliesCount && `(${totalRepliesCount})`}
+        </Text>
+      </Pressable>
+      {isLoading && <ActivityIndicator size="small" color={theme.colors.object.high} />}
```
anicioalexandre: will this change the width of the button? would be nice to make sure it doesnt so we dont have a UI spike during loading. Also, couldnt we use the [Button from DS](https://github.com/silverlogic/baseapp-frontend/blob/master/packages/design-system/components/native/buttons/Button/index.tsx)?

packages/components/modules/comments/native/CommentsList/CommentHideRepliesButton/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { createStyles } from './styles'
+import { CommentHideRepliesButtonProps } from './types'
+
+const CommentHideRepliesButton: FC<CommentHideRepliesButtonProps> = ({ onHideReplies }) => {
+  const theme = useTheme()
+  const styles = createStyles(theme)
+
+  return (
+    <View style={styles.hideRepliesButtonContainer}>
+      <Pressable onPress={onHideReplies} style={styles.hideRepliesButton}>
```
anicioalexandre: same here, can we try to reuse the [Button from DS](https://github.com/silverlogic/baseapp-frontend/blob/master/packages/design-system/components/native/buttons/Button/index.tsx)? so we avoid creating the same classes for a text button variant

packages/components/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # @baseapp-frontend/components
 
+## 3.0.0
```
anicioalexandre: it says minor changes but still bumped into a major, can we double check? sometimes changeset is a bit buggy

anicioalexandre (changes requested):

packages/components/modules/comments/native/CommentDeleteDialog/index.tsx
nossila: Done in f132fe26 — extracted it as `ConfirmDialog` under `packages/design-system/components/native/dialogs/ConfirmDialog`, mirroring the web `ConfirmDialog` API (`title`, `content`, `cancelText`, `action` slot) plus the native close-X title row. Cancel button is built in; `CommentDeleteDialog` is now a thin consumer that only keeps the delete mutation wiring.

packages/components/modules/comments/native/CommentShowRepliesButton/index.tsx
nossila: Done in 3e4a54be — swapped to the DS `Button` (`mode="text" color="inherit" size="small" compact`), which already ships the `buttonSmall` typography this was hand-rolling. On the width concern: the spinner was already rendered outside the button so the label never moved, but the row still grew when it appeared — it now renders inside an always-mounted fixed-size slot, so the layout is identical whether loading or not (I avoided Paper's `loading` prop on purpose, since it injects an inline spinner and shifts the label).

packages/components/modules/comments/native/CommentsList/CommentHideRepliesButton/index.tsx
nossila: Done in 3e4a54be — same DS `Button` text variant here, hand-rolled classes removed.

packages/components/CHANGELOG.md
nossila: Double-checked — it's changesets working as configured, not a glitch. The epic's changeset for `components` is **minor**, but `design-system` is in `components`' (and `wagtail`'s) `peerDependencies`, and with the default `onlyUpdatePeerDependentsWhenOutOfRange: false` changesets major-bumps any dependent whose peer dep is bumped at all. So design-system 2.0.1 → 2.1.0 (minor) mechanically pushed components 2.0.3 → 3.0.0 and wagtail → 3.0.0. There's precedent in the repo: components 2.0.0 is a major whose changelog only shows "Patch Changes" — same mechanism. For consumers the epic is additive (the removed native props were introduced within this same epic branch, never released), so the entry content is correctly "Minor Changes"; the major is purely the peer-bump policy. If we'd rather not major on every peer minor, we could set `onlyUpdatePeerDependentsWhenOutOfRange: true` in `.changeset/config.json`, but that changes release semantics repo-wide so I left it alone.

packages/components/modules/__shared__/native/Timestamp/index.tsx
nossila: Agreed — the TODO right there tracks it. We'll take the `formatDateWithDiffNow` / `formatRelativeTeam` question to the design team as a follow-up so web/native timestamp copy can converge; keeping it out of this epic since it changes visible copy.

anicioalexandre (approved):

packages/components/CHANGELOG.md
nossila: Took a closer look and fixed it properly in f000513c — you're right that we don't need to touch `.changeset/config.json`. The majors were changesets' peer-dependents policy kicking in (design-system is a peerDependency of components/wagtail), not actual breaking changes, and since this branch hand-maintains the version commit anyway, I set the versions the changelog content actually describes: **components 2.0.3 → 2.1.0** (the minor the changeset declared, so the "Minor Changes" heading now matches), **wagtail 2.0.1 → 2.0.2** (dependency-update patch, matching its "Patch Changes" entry), design-system stays 2.1.0. Neither version collides with anything previously released, and since no changeset files remain on the branch, the release action will just publish these versions after merge.

PR #394, opened 2026-06-23
Use profile path on comments redirect url

Ronan-Fernandes:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed profile URL path formatting in comment items to ensure proper path normalization and consistent profile link navigation.

### DIFF
--- a/packages/components/modules/comments/web/CommentItem/index.tsx
+++ b/packages/components/modules/comments/web/CommentItem/index.tsx
@@ -4,6 +4,7 @@ import { FC, useRef, useState, useTransition } from 'react'
 
 import { ClickableAvatar } from '@baseapp-frontend/design-system/components/web/avatars'
 import { Markdown } from '@baseapp-frontend/design-system/components/web/markdown'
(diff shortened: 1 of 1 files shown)


packages/components/modules/comments/web/CommentItem/index.tsx
```diff
@@ -74,7 +75,9 @@ const CommentItem: FC<CommentItemProps> = ({
 
   const profileUrlPath = comment?.profile?.urlPath?.path
   const profileUrl =
-    !profileUrlPath || useProfileId ? `${profilePath}/${comment?.profile?.id}` : profileUrlPath
+    !profileUrlPath || useProfileId
+      ? `${profilePath}/${comment?.profile?.id}`
+      : `${profilePath}/${removeLeadingSlash(profileUrlPath)}`
```
nossila, line 80: this looks wrong, because the profile URL can be either `/profile/:id` or `/username`, when you have the urlPath set for a profile, then it will not start with `/profile/`

nossila (changes requested):

PR #396, opened 2026-06-30
Profile member invitations: invite dialog, member search & expiration handling

nossila:
## Summary

Adds the web UI for the multiple-profiles **member invitation lifecycle** to `@baseapp-frontend/components` (`modules/profiles`), releasing **components 1.8.0**. Two feature groups:

### 1. Member invite dialog
- New **"Add Member"** flow: a dialog with an `allProfiles` typeahead (results in a portaled Popper) plus removable chips for selected profiles and free-text emails.
- Existing profiles are added via `profileUserRoleCreate(usersIds)`; typed emails go through `profileSendInvitation`.
- Gated on the new `canAddMember` permission.

### 2. Expired-invitation handling
- `MemberItemFragment` extended with `invitedEmail` and `invitationExpiresAt`.
- New `ResendInvitation` (`profileResendInvitation`) and `CancelInvitation` (`profileCancelInvitation`) mutations.
- Expired invitations now render in the members list — including email-only invites with no account — with an **"Expired"** label and an action dropdown (**Resend Invitation** / **Remove**); dialogs lifted to fragment level.

## Changes
- **GraphQL** — new mutations `SendInvitation`, `ResendInvitation`, `CancelInvitation`, `ProfileUserRoleCreate`; new `InviteMembersSearch` fragment; updated `MemberItem` / `UserMembersList` fragments; regenerated `schema.graphql`.
- **Components** — new `InviteMemberDialog` (with a `MemberSearch` subcomponent: typeahead + email chips); reworked `MemberItem`, `MemberListItem`, `MembersList`; new `ProfileMembers` constants.
- **Release** — `@baseapp-frontend/components` 1.7.0 → 1.8.0 (CHANGELOG updated).

Scope is **web only** (`modules/profiles/web`); ~1.2k insertions across 22 files.

**Related BE PR:** https://github.com/silverlogic/baseapp-backend/pull/382 — *Multiple Profiles: member management, search & invitation lifecycle* (pinned into the deployable backend via https://bitbucket.org/silverlogic/baseapp-backend-template/pull-requests/286).

## ⚠️ REQUIRES BACKEND WORK
The GraphQL this PR consumes (`profileSendInvitation`, `profileResendInvitation`, `profileCancelInvitation`, `profileUserRoleCreate`, `invitedEmail`, `invitationExpiresAt`, `canAddMember`) ships in the backend PR above, which is **not yet merged** (currently gated on its own dependency). Do not treat this PR as merge-ready until the backend lands.

https://claude.ai/code/session_011PHtxvxF3LJPxhmKtfCEcX

## Summary by CodeRabbit

* **New Features**
  * Added an improved “Add Member” invitation flow with profile typeahead search, selectable chips, and email-based invites with role assignment.
  * Added invitation management, including resend/remove and a dedicated expired-invitation experience with resend/remove actions when permitted.
  * Introduced a new design-system `AutocompleteField` with loading support and debounced input handling.
* **Bug Fixes**
  * Member lists update more reliably after invite actions (including resend/cancel).
  * Member search now falls back to invited email when no profile exists, and the displayed results count better matches searchable content.


nossila:
@coderabbitai review

nossila:
@coderabbitai review

nossila:
@coderabbitai review

nossila:
### CodeRabbit follow-up

All 10 inline review comments from the two CodeRabbit passes were already resolved in earlier commits on this branch (`3e80dbab`, `ea745745`) — verified the code reflects each fix (userId-based dedupe key, `getMutationErrorMessage` empty-message fallback, barrel re-exports, case-insensitive header count, the expiry re-render timer using a fresh `Date.now()`, the WAI-ARIA combobox/listbox keyboard nav in `MemberSearch`, etc.).

The only items still open were the **2 "outside diff range" comments** from the second pass, which were never posted as inline threads (platform limitation) and so weren't picked up with the rest. Both are now addressed in `e265ebdd`:

1. **`InviteMemberDialog` — close while submitting** (`index.tsx`): `handleClose` now returns early while `isSubmitting`, so an in-flight `handleSubmit` continuation can't `resetState()`/`onClose()` a reopened dialog. `isSubmitting` always clears once the batch settles, so the dialog is never trapped open.
2. **`MemberItem` — expired chip a11y** (`index.tsx`): the read-only "Expired" chip now uses `disabled` instead of `sx={{ pointerEvents: 'none' }}`, removing it from the tab order and the a11y tree as an actionable button. The active-role chip is intentionally left as-is — greying out an active role would be misleading.

Separately, `07c311eb` fixes the failing **Component Test Packages** check: the owner-row change in `2e1d0f11` reads `data.owner.profile`, but the `ProfileMembers` test mocks weren't updated to provide it, so the owner row rendered empty and the header undercounted. The mocks now include `owner { id, profile { ...ProfileItemFragment } }`.

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    const profileMembers = selected.filter(isSelectedProfile)
+    const emailMembers = selected.filter(isSelectedEmail)
+
+    const tasks: Promise<SelectedMember[]>[] = []
+    if (profileMembers.length > 0) tasks.push(addMembers(profileId, profileMembers))
+    emailMembers.forEach((member) => tasks.push(inviteEmail(profileId, member)))
+
+    setIsSubmitting(true)
+    const results = await Promise.allSettled(tasks)
```
anicioalexandre, line 119: just wondering what drove multiple single requests vs doing a batch write on the backend? are we limiting the number of invites per time? if we invite like 100 people, wouldn't this break a little sweat?

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const failed = results.filter(isRejected).map((result) => result.reason as BatchError)
+    const failedMembers = failed.flatMap((error) => error.members)
+
+    // Surface each failure (network errors are also toasted by the mutation hooks, but the
+    // batched/field messages here are more specific).
+    failed.forEach((error) => sendToast(error.message, { type: 'error' }))
+
+    // Refetch after ANY successful write so the list reflects the partial batch.
+    if (succeeded.length > 0) {
+      sendToast(succeeded.length === 1 ? 'Member added' : `${succeeded.length} members added`, {
```
anicioalexandre, line 130: not sure about the UX here, when doing around 10, that takes up like 1/4 of the screen. Maybe we could double-check with the design on that?

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+export const isSelectedEmail = (member: SelectedMember): member is SelectedEmail =>
+  member.kind === 'email'
+
+/**
+ * Resolve the first user-facing error message from a mutation's `onCompleted` callback,
+ * checking both the payload's field errors and Relay's top-level transport errors.
+ * Returns `undefined` when the mutation succeeded. Extracted from the dialog's submit
+ * flow so the `onCompleted` handlers stay flat (avoids deeply nested callbacks).
+ */
+export const getMutationErrorMessage = (
```
anicioalexandre, line 24: do you think this can be reused? I think we always do multiple implementations to handle relay errors, but none gets to be reused or catches all the cases

anicioalexandre (approved):

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+      onClose={handleClose}
+      cancelText={COPY.cancel}
+      customMaxWidth={400}
+      content={
+        <Box sx={{ display: 'grid', gap: 2 }}>
+          <Typography variant="body2" color="text.secondary">
+            {COPY.description}
+          </Typography>
+          {open && (
+            <Suspense fallback={<LoadingState CircularProgressProps={{ size: 20 }} />}>
```
anicioalexandre: would be nice if we used `MemberSearch` itself, or a `Textfield` as the fallback so the UI looks a bit more fluid when you open the dialog. Even though it's not typable (can even disable it) it's nice to see the same or a similar element after the one we need lods

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
nossila: Good call — changed it to a batch write. `profileSendInvitation` now takes `emails: [String!]!` (backend PR silverlogic/baseapp-backend#382, `091ffe0`) and the dialog sends every email in a single request, so it's now 2 requests total (one `profileUserRoleCreate` for existing users + one `profileSendInvitation` for emails) regardless of how many you invite. Inviting 100 people is one request now, not 100.

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
nossila: Agreed. Capped the selected-members chip list at `maxHeight: 240` with `overflowY: auto` so a large selection scrolls instead of growing the dialog. Left the exact height as a sensible default — worth confirming the number with design, but it no longer takes over the screen.

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/utils.ts
nossila: Fair point — relay error extraction gets reimplemented a lot. I'd like to do this properly as its own change: promote a single `getMutationErrorMessage`-style helper into `@baseapp-frontend/utils` (covering payload `errors[].messages`, top-level transport errors, and empty-message fallback) and migrate the existing one-offs to it, rather than half-move it here. Opening a follow-up for that so it can catch all the cases in one place.

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/index.tsx
nossila: Done — the Suspense fallback is now a disabled search-style `TextField` (same placeholder) instead of a centered spinner, so the dialog keeps its shape and the real `MemberSearch` input swaps in smoothly once the query resolves.

packages/components/modules/profiles/web/ProfileMembers/InviteMemberDialog/utils.ts
nossila: @anicioalexandre https://app.approvd.io/silverlogic/BA/stories/47006

PR #405, opened 2026-07-14
web add contact to multiple groups

Ronan-Fernandes:
## Summary by CodeRabbit

- **New Features**
  - Added an “Add contact to group” flow from a profile, with group search, selection, and infinite scrolling.
  - Groups the contact already belongs to are disabled and labeled “Already added.”
  - Added confirmation handling with success and error feedback.
- **Bug Fixes**
  - Prevented selection changes for groups where the contact is already a member.
- **Tests**
  - Added coverage for toggling group selection without mutating the previous selection set.


lokmanliton (approved):

packages/components/modules/messages/web/AddContactToGroupDialog/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      DialogContentProps={{
+        sx: {
+          padding: 0,
+          typography: 'body1',
+          color: 'text.primary',
+        },
+      }}
+      content={
+        open ? (
+          <Suspense
```
anicioalexandre: we probably should export GroupsList as a suspended compoenent instead

packages/components/modules/messages/web/AddContactToGroupDialog/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      customMaxWidth={480}
+      DialogContentProps={{
+        sx: {
+          padding: 0,
+          typography: 'body1',
+          color: 'text.primary',
+        },
+      }}
+      content={
+        open ? (
```
anicioalexandre: if we dont want anything to render while the open is false we could also do `if (!open) return null ` or when rendering this component  `open && <AddContactToGroupDialog />`

packages/components/modules/messages/web/AddContactToGroupDialog/GroupsList/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+        <Searchbar
+          name="search"
+          onChange={handleSearchChange}
+          onClear={handleSearchClear}
+          control={searchControl}
+          isPending={isPending}
+        />
+      </SearchbarContainer>
+      {edges.length === 0 ? (
+        renderEmptyState()
```
anicioalexandre: we could probably have just one renderContent() that would return the empty states and at the end return the virtuoso list, that would avoid the ternary

anicioalexandre (approved):

PR #408, opened 2026-07-17
Convert tailwind to MUI + styled

Ronan-Fernandes:


packages/components/modules/messages/web/MessagesList/MessagesGroup/UserMessage/MessageItem/styled.tsx
```diff
@@ -23,3 +25,12 @@ export const MessageItemContainer = styled(Box, {
     },
   }
 })
+
+export const MessageContent = styled(Markdown, {
```
anicioalexandre, line 29: can we double check if this works? i lloks like we need to bump a version to make sure the styles that go into the styled can be correctly spread
> styled(Markdown) works by a fragile mechanism (chat messages). MessageContent = styled(Markdown) relies on Markdown spreading className into ReactMarkdown via ...rest, and react-markdown 9.1.0 wrapping the content in an extra inner <div className>. I verified: both lockfiles pin 9.1.0 ✓, and StyledMarkdown uses only descendant/sibling selectors so the extra div breaks nothing ✓ — rendering is equivalent today. But the catalog range is ^9.0.1, and in 9.0.x className is silently ignored — a future re-resolve would silently drop the message color (other users' messages would lose primary.contrastText). Suggest bumping the catalog to ^9.1.0, or better, having Markdown merge className onto StyledMarkdown explicitly.

packages/components/modules/notifications/web/NotificationsPopover/index.tsx
anicioalexandre: missing the conversion here, we can probably reuse NotificationsButton

anicioalexandre (changes requested):

anicioalexandre:
@Ronan-Fernandes lets also rebase and fix the ci 🙏🏽

PR #409, opened 2026-07-18
fix: prevent sending messages that contain only whitespace (epic integration)

nossila:
## Summary

Epic-integration variant of #406 — the same whitespace-only message fix rebased onto `epic/BA-2239-react-native-comments` as a single commit, so the epic ships with it.

Users could send a message containing only spaces, which appeared as a blank bubble in the conversation thread.

### Root cause

- **Web**: the message input is an MDXEditor (markdown). A spaces-only input is serialized with markdown character references (e.g. `&#x20;`), which survive the schema's `.trim().min(1)` check — so the send button enabled and the "blank" message was sent.
- **Native**: `MessageCreate` invoked its submit handler directly with a `form.watch` value, bypassing zod validation entirely.

### Fix

- Add `hasVisibleContent` (treats `&#x20;`/`&#32;`/`&nbsp;` as whitespace) and use it as a `.refine` on `SOCIAL_UPSERT_FORM_VALIDATION_SCHEMA`. This disables the send button and blocks submission for whitespace-only input on web and native — including the epic's new native comment composer, which shares the schema.
- Route the native `MessageCreate` submit through `form.handleSubmit(onSubmit)` so validation runs at submit time (subsumes the epic's `!body` guard).

### Tests

- New unit tests for `hasVisibleContent` and the schema — full components unit suite green (18/18 on the merged tree), `tsc --noEmit` and eslint clean.

### Relationship to #406

#406 delivers the same fix straight to `master`. This PR delivers it to the epic so the epic branch doesn't regress it; whichever merges second will see the changes already applied (identical content, conflict-free or empty diff after the other lands).

### Note for epic owners

The epic's new shared-native `useCommentComposer` still calls submit directly with a watched value (same pattern messages had). The schema refine keeps its send button disabled for whitespace-only input, but you may want the same `form.handleSubmit` treatment there for defense in depth.

https://claude.ai/code/session_01SHvCU8JqgykLGxNL3FSTFh

### DIFF
--- a/.changeset/tidy-spaces-block.md
+++ b/.changeset/tidy-spaces-block.md
@@ -1,0 +1,5 @@
+---
+"@baseapp-frontend/components": patch
+---
+
+Prevent sending messages that contain only whitespace: reject markdown-escaped whitespace-only bodies (e.g. `&#x20;`) in the social upsert validation schema and run validation on the native MessageCreate submit
--- a/packages/components/modules/__shared__/common/__tests__/socialUpsertFormValidation.test.ts
+++ b/packages/components/modules/__shared__/common/__tests__/socialUpsertFormValidation.test.ts
@@ -1,0 +1,46 @@
+import { SOCIAL_UPSERT_FORM_VALIDATION_SCHEMA } from '../constants'
+import { hasVisibleContent } from '../utils'
+
+const parseBody = (body: string) =>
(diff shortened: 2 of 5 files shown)


nossila:
merged in https://github.com/silverlogic/baseapp-frontend/pull/406

PR #410, opened 2026-07-20
Fix Dropzone remove (×) button re-opening the file picker

nossila:
## Problem

In the New Post composer, after attaching multiple images, clicking the close/remove (×) icon on an attached image preview removed the image but also re-opened the OS file picker.

## Root cause

In `Dropzone`'s multi-file gallery branch, the preview items render inside the container that spreads react-dropzone's `getRootProps()`, whose `onClick` opens the file picker. The remove (×) button's click handler never stopped propagation, so the click bubbled up to the dropzone root and re-triggered the picker. Clicking the image preview itself had the same problem.

Additionally, `RemoveFileButton` is a bare `styled('button')` with no explicit `type`, so it defaulted to `type="submit"` and could submit an enclosing form (e.g. `PostForm`).

## Fix

In `DropzonePreview`:
- Call `event.stopPropagation()` in both the remove (×) button and the image-preview button click handlers.
- Add an explicit `type="button"` to `RemoveFileButton`.

Drag-and-drop over the gallery and the add (+) tile behavior are unchanged. The single-file branch was never affected (its preview doesn't render under `getRootProps`).

Includes a `patch` changeset for `@baseapp-frontend/design-system`.

## How to test

1. Open the New Post form and attach multiple images.
2. Click the × on any preview — only that image is removed; no file picker opens and the form is not submitted.
3. Click a preview thumbnail — it selects/previews the image without opening the picker.
4. Click the add (+) tile — the file picker still opens.

https://claude.ai/code/session_01YGE2YoUpkL4EGXELRKqmTV

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Dropzone preview and remove-file interactions so clicks no longer reopen the file picker.
  * Prevented remove actions from unintentionally submitting forms.

* **Security**
  * Updated the HTTP request library to address known security vulnerabilities.

* **Chores**
  * Released patch updates across authentication, components, design system, GraphQL, provider, utilities, and Wagtail packages.


matheusysd (approved):
I ran coderabbit locally here and the only find is this

> minor [Security & Privacy]
> 
> pnpm-workspace.yaml:37 — "Bump the Axios catalog pin to 1.18.1"
> 
> ▎ axios is pinned to 1.18.0; 1.18.1 is available — a patch release, so update the catalog to keep the patch line current.

overall looks good to me

